PR #63, opened 2024-05-13
fix: fix how pending is calculated

jrwbabylonchain:
- changed API field name from `unconfirmed_active_tvl` to `unconfirmed_tvl`
- fix how pending delegation is calculated as the unconfirmed tvl is the sum of both confirmed and unconfirmed.

### DIFF
--- a/src/app/components/Stats/Stats.tsx
+++ b/src/app/components/Stats/Stats.tsx
@@ -44,10 +44,10 @@ export const Stats: React.FC<StatsProps> = ({
       },
       {
         title: "Pending Stake",
-        value: data?.unconfirmed_active_tvl
+        value: data?.unconfirmed_tvl
           ? formatter.format(
-              +(data.unconfirmed_active_tvl / 1e8).toFixed(6) as number,
-            )
+            +(data.unconfirmed_tvl - data.active_tvl / 1e8).toFixed(6) as number,
+          )
           : 0,
         icon: pendingStake,
       },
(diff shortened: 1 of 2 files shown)


src/app/components/Stats/Stats.tsx
```diff
@@ -44,10 +44,10 @@ export const Stats: React.FC<StatsProps> = ({
       },
       {
         title: "Pending Stake",
-        value: data?.unconfirmed_active_tvl
+        value: data?.unconfirmed_tvl
           ? formatter.format(
-              +(data.unconfirmed_active_tvl / 1e8).toFixed(6) as number,
-            )
+            +(data.unconfirmed_tvl - data.active_tvl / 1e8).toFixed(6) as number,
```
vitsalis, line 49: It's possible that the unconfirmed TVL is negative, e.g. if in the last unconfirmed blocks we only have unbondings.

jrwbabylonchain: i'm going to put this PR on hold. seems i have completely different understanding of the two new fields recent PR added into the queue

vitsalis: Yep, but `unconfirmed_tvl` can be lower than `active_tvl`, e.g. due to unconfirmed unbondings.

jrwbabylonchain: note: https://babylon-chain.slack.com/archives/C04AF7Q8AAC/p1715606656509559
decision: we leave it with negative valuebut keep the wording "pending TVL"

gbarkhatov (approved):

vitsalis (approved):

PR #71, opened 2024-05-14
fix: signPsbt shall return signed Psbt

jrwbabylonchain:
- Update the signPsbt methods, making sure the method only sign Psbt.
- Updated the doc for the missing new method getBTCTipHeight

### DIFF
--- a/src/app/components/Delegations/Delegations.tsx
+++ b/src/app/components/Delegations/Delegations.tsx
@@ -116,9 +116,13 @@ export const Delegations: React.FC<DelegationsProps> = ({
     );
 
     // Sign the unbonding transaction
-    const signedPsbt = await signPsbt(unsignedUnbondingTx.toHex());
-    const unbondingTx = Psbt.fromHex(signedPsbt).extractTransaction();
-
+    let unbondingTx: Transaction;
+    try {
+      const signedPsbt = await signPsbt(unsignedUnbondingTx.toHex());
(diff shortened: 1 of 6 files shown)


src/app/components/Delegations/Delegations.tsx
gbarkhatov, line 1: We need to use the new stuff in the withdrawal method (line `222`)
```ts
// Sign the withdrawal transaction
const signedPsbt = await signPsbt(unsignedWithdrawalTx.toHex());
const withdrawalTransaction = Psbt.fromHex(signedPsbt).extractTransaction();
```

gbarkhatov (changes requested):

src/utils/wallet/okx_wallet.ts
```diff
@@ -87,8 +87,7 @@ export class OKXWallet extends WalletProvider {
       throw new Error("OKX Wallet not connected");
     }
     // sign the PSBT
-    const response = await window?.okxwallet?.bitcoinSignet?.signPsbt(psbtHex);
-    return Psbt.fromHex(response).extractTransaction().toHex();
+    return await window?.okxwallet?.bitcoinSignet?.signPsbt(psbtHex);
```
vitsalis, line 90: We don't need the `Psbt` import anymore

gbarkhatov (approved):

Issue #77, opened 2024-05-15
Internal satoshiToBTC and btcToSatoshi


gbarkhatov:
We should create a 2 separate functions that convert:
1. satoshis to btc
2. btc to satoshis

Satoshis should be used everywhere, while BTC is only for the presentation layer

---

Should we pass the amount in satoshis? I would recommend renaming this to `amountBTC` in the react component and converting it to `amount` before passing it around, i.e. let's let this component only have to care about display conversions of amount, while we use satoshis everywhere else.

In general, I recommend that internally all of our state should only be in Satoshis and we use BTC only for input/output in terms of presentation. Anything we pass around in the application internally should be in satoshis.

_Originally posted by @vitsalis in https://github.com/babylonchain/simple-staking/pull/74#discussion_r1601989863_

Issue #21, opened 2024-04-04
Types refactoring


gbarkhatov:
Ideally, we would have API response types and internal types (which contain only the things we need).
The API response types need to be converted into the internal types of the application, so that internal methods do not depend on the data format of the API. We can do this in a future PR.
We have duplication structures in api, we can move them to src/types or api/types.ts

PR #80, opened 2024-05-16
Correct using btc height for params

jrwbabylonchain:
- Added loading state which will wait for the params to be loaded
image

- When current height + 1 is < than first activation height
image

- When getting close to the next version
image


jrwbabylonchain:
This PR does not deal with the UX representation of each scenarios. It shall be designed first, then apply with the appropriate languages.

src/utils/signForm.ts
```diff
@@ -0,0 +1,86 @@
...
+import { Psbt, networks } from "bitcoinjs-lib";
+import { apiDataToStakingScripts } from "./apiDataToStakingScripts";
+import { stakingTransaction } from "btc-staking-ts";
+import { isTaproot } from "./wallet";
+
+export const signForm = async (
+    params: GlobalParamsVersion,
+    btcWallet: WalletProvider,
+    finalityProvider: FinalityProvider,
+    stakingTerm: number,
```
vitsalis, line 13:
```suggestion
    stakingTime: number,
```

src/app/page.tsx
```diff
@@ -171,126 +173,58 @@ const Home: React.FC<HomeProps> = () => {
...
+        throw new Error("Wallet network not connected");
+      } else if (!paramWithContext || !paramWithContext.currentVersion) {
+        throw new Error("Global params not loaded");
+      } else if (!finalityProvider) {
+        throw new Error("Finality provider not selected");
+      }
+      const { currentVersion: globalParamsVersion } = paramWithContext;
+      // Rounding the input since 0.0006 * 1e8 is is 59999.999
+      // which won't be accepted by the mempool API
+      const stakingAmount = Math.round(Number(amount) * 1e8);
```
vitsalis, line 190: Instead of the staking conversion happening here we should have an `onChange()` that stores `amount` as the satoshis value. #77 can be done in another PR.

src/app/page.tsx
```diff
@@ -171,126 +173,58 @@ const Home: React.FC<HomeProps> = () => {
...
+      const stakingTerm = getStakingTerm(globalParamsVersion, term);
```
vitsalis, line 191:
```suggestion
      const stakingTime = getStakingTime(globalParamsVersion, stakingTime);
```

src/app/page.tsx
```diff
@@ -171,126 +173,58 @@ const Home: React.FC<HomeProps> = () => {
...
+      ]);
 
-    let txID;
-    try {
-      txID = await btcWallet.pushTx(stakingTx);
     } catch (error: Error | any) {
-      console.error(error?.message || "Broadcasting staking transaction error");
+      // TODO Show Popup
+      console.error(error?.message || "Error signing the form");
     }
```
vitsalis, line 226: Don't like wrapping the whole transaction in a try catch, leaves us with little control. I would prefer if we put the logic inside the `try` in a separate function and we just call that from here. If we have to wrap all this logic inside a `try/catch` then this logic inside the `try` deserves its own function.

src/utils/globalParams.ts
```diff
@@ -0,0 +1,73 @@
...
+  const [height, versions] = await Promise.all([getCurrentBtcHeight(), getGlobalParams()])
+
+  // sorted in descending order
+  const sorted = versions.sort(
+    (a, b) => b.activationHeight - a.activationHeight,
+  );
+
+  for (let i = 0; i < sorted.length; i++) {
+    const curr = sorted[i];
+    if (curr.activationHeight <= height + 1) {
```
vitsalis, line 52: Better for the caller to specify the height instead of us being smart here.

src/utils/globalParams.ts
```diff
@@ -0,0 +1,73 @@
+import {
+  GlobalParamsVersion,
+  getGlobalParams,
+} from "@/app/api/getGlobalParams";
+
+
+export const getGlobalParamsVersioByHeight = async (
```
gbarkhatov, line 7: Versio -> Version

src/utils/globalParams.ts
```diff
@@ -0,0 +1,73 @@
+import {
+  GlobalParamsVersion,
+  getGlobalParams,
+} from "@/app/api/getGlobalParams";
+
+
```
gbarkhatov, line 6: Should we add a comment about what this function does?

src/utils/globalParams.ts
gbarkhatov: At the moment it's somewhat hard to understand what's going on, especially with an imperative style. If this (current) approach is taken I would suggest commenting on what we're doing there

src/utils/globalParams.ts
```diff
@@ -0,0 +1,73 @@
...
+        };
+      }
+      return {
+        currentVersion: curr,
+        isApprochingNextVersion: false,
+      };
+    }
+  }
+
+  return {
```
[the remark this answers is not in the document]
gbarkhatov: Or else condition. Hard to understand with imperative + double `if`s with double `returns`

src/utils/signForm.ts
gbarkhatov: We need to either stick with the `term` or with the `time`, I guess `time` is better there but maybe we are to shift away from the `term` completely? Can be extracted to a separate PR

src/utils/signForm.ts
```diff
@@ -0,0 +1,86 @@
+import { GlobalParamsVersion } from "@/app/api/getGlobalParams";
```
gbarkhatov, line 1: I suggest splitting global packages imports with the local imports
i.e.
```ts
import { Psbt, networks } from "bitcoinjs-lib";
import { stakingTransaction } from "btc-staking-ts";

import { GlobalParamsVersion } from "@/app/api/getGlobalParams";
import { FinalityProvider } from "@/app/api/getFinalityProviders";
import { WalletProvider } from "./wallet/wallet_provider";
import { apiDataToStakingScripts } from "./apiDataToStakingScripts";
import { isTaproot } from "./wallet";
```

gbarkhatov (commented):
@jrwbabylonchain 
1. you need to run `npm run format:fix`, or add proper plugins to VSCode (Eslint, etc.), or we can add this as a hook on each commit
2. I need to be able to run this using the UI, currently I receive `Invalid staking data` https://i.imgur.com/z6fcySY.png

src/app/components/Staking/Staking.tsx
```diff
@@ -31,43 +38,28 @@ export const Staking: React.FC<StakingProps> = ({
...
-    maxAmountBTC &&
-    amount >= minAmountBTC &&
-    amount <= maxAmountBTC;
-
-  const signReady = amountReady && termsReady && selectedFinalityProvider;
 
-  const renderFixedTerm = () => {
-    if (minTerm && maxTerm && minTerm === maxTerm) {
+  const renderFixedTerm = (isDisabled: boolean, params: StakingParams) => {
+    const { minStakingTime, maxStakingTime } = params;
```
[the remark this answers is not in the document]
jrwbabylonchain: The method itself defines what value is acceptable. `undefined` is not allowed here in the param. If you talking about whether we safety check the undefined value before calling this method, `undefined` value it's checked on line
```
// 3. wallet is connected but staking has not started
    else if (!stakingParams) {
```

src/app/components/Staking/Staking.tsx
```diff
@@ -89,14 +81,14 @@ export const Staking: React.FC<StakingProps> = ({
             type="number"
```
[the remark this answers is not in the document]
jrwbabylonchain: This would violating the type checking at compile time, no? It shall never happen IMO as the param is already defined as accepting input being a `StakingParams` type.

src/app/components/Staking/Staking.tsx
```diff
@@ -133,8 +163,23 @@ export const Staking: React.FC<StakingProps> = ({
...
       );
     }
-    // 3. wallet is connected and staking cap is not reached
     else {
+      const { minStakingTime, maxStakingTime, minStakingAmount, maxStakingAmount } = stakingParams;
+      const termsReady =
+        (minStakingTime === maxStakingTime || (term >= minStakingTime && term <= maxStakingTime));
+
+      const minAmountBTC = minStakingAmount ? minStakingAmount / 1e8 : 0;
+      const maxAmountBTC = maxStakingAmount ? maxStakingAmount / 1e8 : 0;
```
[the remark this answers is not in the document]
jrwbabylonchain: yeah, i do not intend to change existing logic. a lot of code are copied over in this PR as what it is before (moved location). we shall do refactoring piece by piece in seperate PRs.

src/app/components/Staking/Staking.tsx
```diff
@@ -133,8 +163,23 @@ export const Staking: React.FC<StakingProps> = ({
         </div>
       );
     }
-    // 3. wallet is connected and staking cap is not reached
     else {
+      const { minStakingTime, maxStakingTime, minStakingAmount, maxStakingAmount } = stakingParams;
+      const termsReady =
+        (minStakingTime === maxStakingTime || (term >= minStakingTime && term <= maxStakingTime));
```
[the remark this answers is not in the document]
jrwbabylonchain: this code was not changed, just moved location. 
The `term` was used quite widely in the repo including in the UX text. I think change it to `stakingTime` might not be better than keep it as it is.
i let @gbarkhatov decide on the name of the variables here. i don't mind at all.

src/app/components/Staking/Staking.tsx
```diff
@@ -143,7 +188,7 @@ export const Staking: React.FC<StakingProps> = ({
           </div>
           <div className="flex flex-1 flex-col">
             <div className="flex flex-1 flex-col">
-              {renderFixedTerm()}
+              {renderFixedTerm(!isWalletConnected, stakingParams)}
```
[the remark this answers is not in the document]
jrwbabylonchain: I agree on this.
what about the ones below? since we already connect to the wallet. why are we still checking wallet connection to disable the form?

src/utils/globalParams.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  const versions = await getGlobalParams();
+
+  // sorted in descending order
+  const sorted = versions.sort(
+    (a, b) => b.activationHeight - a.activationHeight,
+  );
+
+  // find the first version that is less than or equal to the current height + 1 (next block height)
```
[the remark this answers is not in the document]
jrwbabylonchain: ah, yes. the logic is written without the `+ 1`. but forgot to remove/update this comment.

src/utils/globalParams.ts
```diff
@@ -0,0 +1,73 @@
...
+    (v) => v.activationHeight <= height,
+  );
+
+  if (!params) {
+    throw new Error("No global params version found");
+  }
+  return params;
+};
+
+export interface GlobalParamsWithContext {
```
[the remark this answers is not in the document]
jrwbabylonchain: There is no `GlobalParams` type anywhere tho.
Also typescript is different to go, you don't need to define the public types under a different directory to avoid circular dependency. The type here is just to assistant with making the method definition looks shorter.
I think what's best can be done here is not to make it public as we don't need to use this type anywhere.

src/utils/signForm.ts
jrwbabylonchain: agree. also commented on other comment about this. Let's not worry about `term` vs `time` here. `term` has been widely used in this repo, so if we wanna change it to `time`, better do it in a single PR to rename all.

src/utils/globalParams.ts
```diff
@@ -0,0 +1,73 @@
+import {
+  GlobalParamsVersion,
```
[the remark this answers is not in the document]
jrwbabylonchain: It's not necessary in typescript as there is no issue of circular dependency like in Go. I have only seen exposing interfaces under a dedicated type directory for libs that want to expose it publicly.
For internal types coming from the API response (after we coverted it into our own types), it shall be common to use it directly in the repo.

What shall be done in this file as you have already outlined in other comment is to take out the REST call to API. It doesn't make sense in a util method. I just renamed the file name and extended the methods, did not thinking too much about this haha. I completely agree with you https://github.com/babylonchain/simple-staking/pull/80/files/24e3cfa7e163257bd89c14a8763ce101c2fdd536#diff-5e523d0587392747c08ecc53b3492a937231a00caab19beffb81d2fbe64e1cdfL15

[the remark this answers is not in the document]
vitsalis: I just thought it would be cleaner if we had our internal types in a separate types package instead of importing through the API.

src/utils/globalParams.ts
```diff
@@ -0,0 +1,73 @@
...
+
+
+export const getGlobalParamsVersioByHeight = async (
+  height: number,
+): Promise<GlobalParamsVersion> => {
+  if (!height) {
+    throw new Error("Height is not provided");
+  }
+
+  const versions = await getGlobalParams();
```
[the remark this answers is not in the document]
jrwbabylonchain: Completely agree on this.
I had a quick look into the exist codebase ([the implementation of this method is the same as what it is before](https://github.com/babylonchain/simple-staking/pull/80/files/24e3cfa7e163257bd89c14a8763ce101c2fdd536#diff-5e523d0587392747c08ecc53b3492a937231a00caab19beffb81d2fbe64e1cdfL15)), I would recommend we rethink on how the global params are being passed around through props. Ideally, **the global param API should only be called at very high level of the react flow**. For example, it shall be called at the page.tsx only and passing down through react props to each component. Right now, we are making an API call each time we require this global params, this is not in its best way of handling such usage

To avoid making this PR goes forever and making too many refactoring of the existing codebase, what i will do is only move the `await getGlobalParams();` out of here and call it at the `handleUnbond` instead. This would at least make the util method a legit method.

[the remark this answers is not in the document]
vitsalis: Let's create an issue to track the above comment?

src/utils/globalParams.ts
jrwbabylonchain: added the comment and made the height + 1 being controlled by the calling method.
In general i think the effect is the same, you either make the method name and comment description clear enough so that the calling method is aware the params we getting is for the next block. Or as you suggest let the calling method to make this call. I don't have strong opinion on this.

src/utils/globalParams.ts
jrwbabylonchain: i'm going to remove this method

jrwbabylonchain:
> @jrwbabylonchain
> 
> 1. you need to run `npm run format:fix`, or add proper plugins to VSCode (Eslint, etc.), or we can add this as a hook on each commit
> 2. I need to be able to run this using the UI, currently I receive `Invalid staking data` https://i.imgur.com/z6fcySY.png

Let's add it in the commit hook. i will create an issue for this

src/utils/globalParams.ts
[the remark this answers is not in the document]
jrwbabylonchain: #85

src/app/components/Staking/Staking.tsx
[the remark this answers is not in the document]
gbarkhatov: @jrwbabylonchain Whatever makes sense now, we have extremely tight deadlines with changing stuff, so we can rename our variables after

src/utils/globalParams.ts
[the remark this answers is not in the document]
gbarkhatov: > For example, it shall be called at the page.tsx only and passing down through react props to each component. Right now, we are making an API call each time we require this global params, this is not in its best way of handling such usage

If we use `react-query` and put the same name query in each of the components it won't call the API all the time, it has caching. We can manually set the refetch interval or refetch condition

>  Right now, we are making an API call each time we require this global params, this is not in its best way of handling such usage
Aren't we supposed to have the latest global params when we do the `actions` (stake, unbond, withdraw)? I.e. the global params in page.tsx is 60 seconds, 30 seconds passed and we stake - what if params are changed?

src/app/api/getGlobalParams.ts
```diff
@@ -1,33 +1,62 @@
+import { AxiosResponse } from "axios";
 import { apiWrapper } from "./apiWrapper";
 
-interface GlobalParams {
-  data: GlobalParamsData;
-}
-
-export interface GlobalParamsData {
-  versions: GlobalParamsVersion[];
+interface GlobalParamsDataResponse {
```
vitsalis, line 4: #21 putting this here as we should migrate internal types to a types package.

src/app/components/Staking/Staking.tsx
```diff
@@ -132,9 +162,28 @@ export const Staking: React.FC<StakingProps> = ({
...
+    } else {
+      const {
+        minStakingTime,
+        maxStakingTime,
+        minStakingAmount,
+        maxStakingAmount,
+      } = stakingParams;
+      const stakingTimeReady =
+        minStakingTime === maxStakingTime ||
+        (term >= minStakingTime && term <= maxStakingTime);
```
vitsalis, line 174: Shouldn't the `term >= minStakingTime && term <= maxStakingTime` condition cover it?

src/utils/globalParams.ts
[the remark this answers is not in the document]
jrwbabylonchain: @gbarkhatov i don't think we should worry about the global param being changed. the timeframe from activation height A to activation B suppose to be big. At least in terms of days. 
The fetching interval is in minutes, so i would say we are fine on this. no need to fetch upon each tx submission.

src/utils/getStakingTerm.ts
```diff
@@ -0,0 +1,14 @@
+import { GlobalParamsVersion } from "@/app/api/getGlobalParams";
+
+export const getStakingTerm = (params: GlobalParamsVersion, term: number) => {
```
vitsalis, line 3: I think instead of creating this complex logic, we should just set the staking time in the React component to a static value when the values are the same. Eitherway the form that updates it won't be visible, so we would be good. We can do this in another PR.

src/utils/globalParams.ts
```diff
@@ -0,0 +1,38 @@
+import { GlobalParamsVersion } from "@/app/api/getGlobalParams";
+
+// This function returns the current global params version and whether the next version is approaching
+// The current version is the one that is active at the current height
+// Return currentVersion being undefined if no version is found,
+// which means the current height is lower than the first version
+export const getCurrentGlobalParamsVersion = async (
+  height: number,
+  versionedParams: GlobalParamsVersion[],
+) => {
```
vitsalis, line 10: We have not specified a return type

vitsalis (approved):
lgtm, let's get this out of the way so we can work on next updates

src/utils/getStakingTerm.ts
jrwbabylonchain: yeah, i did not spend much time looking into the logic of term/staking time. Hence no comments on this.
I only moved the logic out of the page.tsx because the file is way too big and this part of logic seems to be a good utils method 😄

src/utils/globalParams.ts
jrwbabylonchain: defining return type is optional in typescript. it will infer it based on the function's return statements. but yeah, specifying it is a good practise haha. next PR. 
(if type mismatch, there will be error at compiling time.)

PR #90, opened 2024-05-18
fix: send lock height to staking tx

jrwbabylonchain:


jrwbabylonchain:
@gbarkhatov do we bump the version in package.json for this repo as well?

vitsalis (approved):
lgtm. Wonder whether we should make the withdraw and unbonding transactions RBF-able as well. Would require an update only on btc-staking-ts

src/utils/signForm.ts
```diff
@@ -67,6 +67,7 @@ export const signForm = async (
       btcWalletNetwork,
       isTaproot(address) ? Buffer.from(publicKeyNoCoord, "hex") : undefined,
       dataEmbedScript,
+      params.activationHeight,
```
vitsalis, line 70: Actually, shouldn't this be `activationHeight - 1`? Given that lockTime is the block * after * which the transaction can be included

vitsalis: @filippos47 in our tests let's verify whether
1. The application opens up for staking 1 height before the first activation height
2. The transactions that are submitted can be included in the activation height itself

jrwbabylonchain:
> lgtm. Wonder whether we should make the withdraw and unbonding transactions RBF-able as well. Would require an update only on btc-staking-ts

I don't have comments on this.
One of the thing i noticed is that we using v0 of the psbt in stakingtx but using v2 in withdraw. I don't know the consequence of this.

src/utils/signForm.ts
jrwbabylonchain: @vitsalis Super good catch. I think you are correct after I digging deep into BTC code https://github.com/bitcoin/bitcoin/blob/058af75874ffa2b4064e3d6d30cc50f0ec754ba8/src/consensus/tx_verify.cpp#L21

gbarkhatov:
> @gbarkhatov do we bump the version in package.json for this repo as well?

we can bump the version once the UI stuff is there (next week)

gbarkhatov (approved):

PR #92, opened 2024-05-18
feature: Staking form redesign

### LINKED ISSUE #87: Term / duration field
- When we have a "regular" form state with min and max term values -> We have to input the staking term to be able to stake and click the button. We have API data as the HTML input constraint and we use internal **react input state value**.
- When we have a "fixed term" (min === max) we don't input the data into the field, because it's hidden. But the form should be able to be submitted and staking transaction to be created, even though we **are relying on API data**

We need to be sure that the code for this part is non-problematic, simple to understand and manage, and that no errors are happening "in-between". This might involve:
- staking form component restructure
- staking logic (`handleSign`) to be checked
- applying all the constraints properly

Keep in mind, that our form is shown when the wallet is not connected

I suggest completing this ticket after the general form validation #76

### LINKED ISSUE #88: New staking form design
When we get the latest staking form design with all its states (loading, upgrade, not activated) we need to implement the changes

gbarkhatov:
Staking form states:

1. Wallet is not connected https://i.imgur.com/GaCgE30.png https://i.imgur.com/GUvpoz3.png
2. Wallet is connected but we are still loading the staking params https://i.imgur.com/u4PVqBb.png https://i.imgur.com/eXBkoQQ.png
3. Staking has not started yet https://i.imgur.com/EVDMLBU.png https://i.imgur.com/6x41c1L.png
4. Staking params upgrading https://i.imgur.com/dghg28U.png https://i.imgur.com/9YkA8sm.png
5. Staking cap reached https://i.imgur.com/fyNq9xa.png https://i.imgur.com/fKDPhgN.png
6. Min and max terms are equal -> fixed term, no term input, amount, finality provider https://i.imgur.com/17JdFA1.png https://i.imgur.com/pmNwMkq.png
7. Regular form -> amount, term, finality provider https://i.imgur.com/162iJnI.png https://i.imgur.com/XlCSgAj.png

Additional UI changes:
- Changed the fonts so weights are working

**Code changes:**
- This is the first ticket that moves some of the logic from the `page.tsx` so it becomes just a router as @supertong suggested -> `<Staking />` component is responsible for `handleSign` and initial staking creation
- I split all the `Staking` form states into different components. Should close #87. These 2 components are opening the confirmation staking modal and is responsible for any further form validations.
  - `<Fixed />` is responsible for fixed term logic and passes through API value, only amount should be set
  - `<Regular />` is a regular form that has both the amount and term inputs
  - Applied the naming of `amountSat` and `termBlock` so it's clearer what data we're using. Should help with #77 
 
The picture is
staking diagram

`<Form /> can either display "messages" (which can't be combined into one component as `WalletNotConnected` looks different in the UI) or display a `<Fixed />` or `<Regular />` form.

Theoretically, we can extract some parts of the UI from `Fixed` and `Regular` to `Form` or even `Staking` since these components have common logic and checks, but then we will have the same problem we had with #87 - one component, that decides whether or not we should use react state check (term) or use an API value.

So with splitting to `Fixed` and `Regular` we have repetitive logic, but only these components are responsible what they will pass inside the confirmation (`<Preview />) modal and `handleSign` eventually. I.e. we're more verbose and repetetive but have more control over how fixed term form or how regular form will look like.

Closes #88


src/app/components/Modals/PreviewModal.tsx
```diff
@@ -3,22 +3,23 @@ import { Modal } from "react-responsive-modal";
...
 
 interface PreviewModalProps {
   open: boolean;
   onClose: (value: boolean) => void;
   onSign: () => void;
   finalityProvider: string | undefined;
-  amount: number;
-  term: number;
+  amountSat: number;
+  termBlocks: number;
```
vitsalis, line 14: How about we use `stakingTimeBlocks`?

src/app/components/Staking/Form/Fixed.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface FixedProps {
+  onSign: (amountSat: number, termBlocks: number) => Promise<void>;
+  selectedFinalityProvider: FinalityProviderInterface | undefined;
+  minStakingAmount: number;
+  maxStakingAmount: number;
+  stakingTime: number;
+}
+
+// Fixed term + amount with fp check
+export const Fixed: React.FC<FixedProps> = ({
```
vitsalis, line 16: Let's use a more descriptive name for this component, e.g. `FixedStakingTime`.

src/app/components/Staking/Form/Regular.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onSign: (amountSat: number, termBlocks: number) => Promise<void>;
+  selectedFinalityProvider: FinalityProviderInterface | undefined;
+  minStakingAmount: number;
+  maxStakingAmount: number;
+  minStakingTime: number;
+  maxStakingTime: number;
+}
+
+// Regular term + amount form with fp check
+export const Regular: React.FC<RegularProps> = ({
```
vitsalis, line 16: `Regular` is not a good name. How about `ModifiableStakingTime` or something similar?

src/app/components/Staking/Form/Regular.tsx
```diff
@@ -0,0 +1,117 @@
...
+  minStakingTime,
+  maxStakingTime,
+}) => {
+  const [previewModalOpen, setPreviewModalOpen] = useState(false);
+  const [amountBTC, setAmountBTC] = useState(minStakingAmount / 1e8);
+  const [termBlocks, setTermBlocks] = useState(minStakingTime);
+
+  // TODO extract BTC <- -> sBTC conversion to a helper function
+  const minAmountBTC = minStakingAmount ? minStakingAmount / 1e8 : 0;
+  const maxAmountBTC = maxStakingAmount ? maxStakingAmount / 1e8 : 0;
```
vitsalis, line 30: This logic is duplicated among `Fixed` and `Regular`.

src/app/components/Staking/Form/Fixed.tsx
```diff
@@ -0,0 +1,106 @@
...
+  onSign,
+  selectedFinalityProvider,
+  minStakingAmount,
+  maxStakingAmount,
```
vitsalis, line 20: To follow the other updates, how about we name those `minStakingAmountSat` and `maxStakingAmountSat`?

src/app/components/Staking/Form/Fixed.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  // TODO extract BTC <- -> sBTC conversion to a helper function
+  const minAmountBTC = minStakingAmount ? minStakingAmount / 1e8 : 0;
+  const maxAmountBTC = maxStakingAmount ? maxStakingAmount / 1e8 : 0;
+
+  const amountReady =
+    minAmountBTC &&
+    maxAmountBTC &&
+    amountBTC >= minAmountBTC &&
+    amountBTC <= maxAmountBTC;
```
vitsalis, line 34: Would do the comparisons in satoshis.

src/app/components/Staking/Form/Fixed.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  // reset the component state
+  const handleSign = () => {
+    setAmountBTC(0);
+    setPreviewModalOpen(false);
+    onSign(amountSat, stakingTime);
+  };
+
+  // Rounding the input since 0.0006 * 1e8 is is 59999.999
+  const amountSat = Math.round(amountBTC * 1e8);
```
vitsalis, line 46: Why not do this conversion at the top?

src/app/components/Staking/Form/Fixed.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <p>Set up staking terms</p>
+      <div className="flex flex-1 flex-col">
+        <div className="flex flex-1 flex-col">
+          <div className="card mb-2 bg-base-200 p-4">
+            <p>
+              Your Signet BTC will be staked for a fixed term of{" "}
+              {blocksToTime(stakingTime)}.
+            </p>
+            <p>
+              But you can unbond and withdraw your Signet BTC anytime through
```
vitsalis, line 59:
```suggestion
             You can unbond and withdraw your Signet BTC anytime through
```

src/app/components/Staking/Form/Fixed.tsx
```diff
@@ -0,0 +1,106 @@
...
+          </label>
+        </div>
+        <button
+          className="btn-primary btn mt-2 w-full"
+          disabled={!signReady}
+          onClick={() => setPreviewModalOpen(true)}
+        >
+          Preview
+        </button>
+        <PreviewModal
```
vitsalis, line 95: It's a bit weird that half of the form is in another component, but the `Preview` button is in a sibling component. Ideally, I would expect that `Form` is a root component that has children related to its fields, which notify `Form` about updates through callbacks (like you do with `FinalityProviders`). None of the children know about the other children. `Preview` is also a child component of `Form` and gets passed the values from the other components.

src/app/components/Staking/Form/Form.tsx
```diff
@@ -0,0 +1,105 @@
...
+      />
+    );
+  }
+  // 5. Staking cap reached
+  else if (overTheCap) {
+    return (
+      <Message
+        title="Staking cap reached"
+        messages={[
+          "Staking is temporarily disabled due to cap reached.",
```
vitsalis, line 72:
```suggestion
          "Staking is temporarily disabled due to the staking cap getting reached.",
```

src/app/components/Staking/Form/Form.tsx
```diff
@@ -0,0 +1,105 @@
...
+          "Please check your staking history to see if any of yourstake is tagged overflow.",
```
vitsalis, line 73:
```suggestion
          "Please check your staking history to see if any of your stake is tagged overflow.",
```

vitsalis (commented):
This is very good progress! First round for now, will review tomorrow again.

I find that the logic is a bit duplicated between `Fixed` and `Regular`. How about we have the following structure: We have a root component (e.g. `StakingForm`), which has the following child components:
- `FinalityProviders` -> Displays the finality providers. They always appear.
- Depending on the conditions, it displays either the disabled messages, or the staking input fields (as you already do in the `Form` component right now):
    - If the form should be disabled, just display `WalletNotConnected`, `Loading`, or `Message`.
    - Otherwise, if the form should be enabled, include the following components:
         - `StakingTerm` -> Handles all the logic related to staking time. It decides what to display depending on whether the term is static or not and lets the upper form know about the final term, e.g. if the term is static, it just invokes the callback once. If the term is not static the callback is invoked every time the input changes.
         - `StakingAmount` -> Displays the staking amount that the user selects.
         - `Preview` -> As we already do


The above architecture ensures the following:
- We have a central root/controller component that controls staking, i.e. the fields that should be collected.
- We isolate the logic of each particular thing we need to collect to its particular component (makes for easier to check code). The fields that are selected are communicated with the root/controller component through callbacks.
- We do not have duplication of code: For example, we do not need to duplicate the selection of the staking amount or the preview modal just because we display different things on fixed staking term vs. variable one. Let the `StakingTerm` component handle this logic and communicate the selected value (or the static one) to the root/controller component.

src/app/components/Staking/Form/Fixed.tsx
supertong, line 95: My 2 cents. Ideally we wouldn't want the leaf component to import a component from a top-level unless the component is from the `common/ui` folder. However, I think what @vitsalis suggested makes more sense if we could lift the `<PreviewModal />` to be inside `<Form />` and this way we don't even need to make `<PreviewModal />` a globally shared component.

src/app/components/Staking/Form/Fixed.tsx
gbarkhatov: Closed in #104

gbarkhatov:
@vitsalis implemented the approach

- Forms now can be properly checked and validated since they are micro components
- Fixed form check happens on StakingForm
- Resolved your requests

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
-  // if the staking cap is reached, the user can't stake
+  overTheCap: boolean;
   onConnect: () => void;
+  btcWallet: WalletProvider | undefined;
+  btcWalletNetwork: networks.Network | undefined;
+  address: string | undefined;
+  publicKeyNoCoord: string;
+  setDelegationsLocalStorage: Dispatch<SetStateAction<Delegation[]>>;
+  paramWithContext:
+    | {
```
vitsalis, line 38: What is the `| ` this line starts with?

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
-  isUpgrading,
+  paramWithContext,
+  btcWallet,
+  btcWalletNetwork,
+  address,
+  publicKeyNoCoord,
+  setDelegationsLocalStorage,
 }) => {
+  // Staking form state
+  const [stakingAmountBTC, setStakingAmountBTC] = useState(0);
```
vitsalis, line 59:
```suggestion
  const [stakingAmountSat, setStakingAmountSat] = useState(0);
```

We should always use satoshis internally.

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
+      throw new Error("Address is not set");
+    } else if (!btcWalletNetwork) {
+      throw new Error("Wallet network not connected");
+    } else if (!paramWithContext || !paramWithContext.currentVersion) {
+      throw new Error("Global params not loaded");
+    } else if (!finalityProvider) {
+      throw new Error("Finality provider not selected");
+    }
+    const { currentVersion: globalParamsVersion } = paramWithContext;
+    const stakingAmountSat = amountSat;
```
vitsalis, line 81: Why are we setting this new variable?

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
-          <div className="label flex justify-end">
-            <span className="label-text-alt">
-              min term is {minStakingTime} blocks
-            </span>
-          </div>
-        </label>
+  const stakingParams = paramWithContext?.currentVersion;
+  const isUpgrading = paramWithContext?.isApprochingNextVersion;
+
+  const handleSign = async (amountSat: number, stakingTimeBlocks: number) => {
```
vitsalis, line 68: This function should be invoked by this component, not by the `Preview` component. The `Preview` modal can tell the `Staking` component when "submit" was clicked, and it invokes this by only using the values that are already in the state, i.e. `amountSat` and `stakingTimeBlocks`

If we rely on getting those from the parameters we open ourselves to issues in which `amountSat` and `stakingTimeBlocks` that are passed are different than the ones that are in the current state.

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
+        signedTxHex,
+        stakingTerm,
+      ),
+      ...delegations,
+    ]);
+
+    setFinalityProvider(undefined);
+    setStakingAmountBTC(0);
+    setStakingTimeBlocks(0);
+    setPreviewModalOpen(false);
```
vitsalis, line 126: We should have a `restState` function that does the above calls.

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
+  };
+
+  // Select the finality provider from the list
+  const handleChooseFinalityProvider = (btcPkHex: string) => {
+    if (!finalityProviders) {
+      return;
+    }
+    const found = finalityProviders.find((fp) => fp?.btcPk === btcPkHex);
+    if (found) {
```
vitsalis, line 135: Shouldn't it be an error if not found?

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
+
+    setFinalityProvider(undefined);
+    setStakingAmountBTC(0);
+    setStakingTimeBlocks(0);
+    setPreviewModalOpen(false);
+  };
+
+  // Select the finality provider from the list
+  const handleChooseFinalityProvider = (btcPkHex: string) => {
+    if (!finalityProviders) {
```
vitsalis, line 131: Shouldn't it be an error if finality providers are not set and one is selected?

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
+    }
+    const found = finalityProviders.find((fp) => fp?.btcPk === btcPkHex);
+    if (found) {
+      setFinalityProvider(found);
     }
   };
 
-  const renderContentForm = () => {
-    // 1. wallet is not connected
+  const handleStakingAmountChange = (inputAmountBTC: number) => {
```
vitsalis, line 140: We should always be using satoshis internally.

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
-      const amountReady =
-        minStakingAmountSat &&
-        maxStakingAmountSat &&
-        amountSat >= minStakingAmountSat &&
-        amountSat <= maxStakingAmountSat;
 
-      const signReady =
-        amountReady && stakingTimeReady && selectedFinalityProvider;
+      // Check if the staking transaction is ready to be signed
+      const signReady = stakingSignReady(
```
vitsalis, line 206:
```suggestion
      const signReady = isStakingSignReady(
```

src/app/components/Staking/Staking.tsx
```diff
@@ -1,225 +1,239 @@
...
+        finalityProvider,
+      );
 
-      // Do the conversion to BTC for presentational purposes
-      const minStakingAmountBtc = satoshiToBtc(minStakingAmountSat);
-      const maxStakingAmountBtc = satoshiToBtc(maxStakingAmountSat);
-      const selectedStakingAmountBtc = satoshiToBtc(amountSat);
+      // Staking time is fixed
+      const stakingTimeFixed = minStakingTimeBlocks === maxStakingTimeBlocks;
+      const amountSat = Math.round(btcToSatoshi(stakingAmountBTC));
```
vitsalis, line 218: Our state should contain the amount in satoshis, not in BTC.

src/utils/stakingSignReady.ts
```diff
@@ -0,0 +1,40 @@
...
+
+// Check if the staking transaction is ready to be signed
+export const stakingSignReady = (
+  // API values
+  minStakingAmountSat: number,
+  maxStakingAmountSat: number,
+  minStakingTimeBlocks: number,
+  maxStakingTimeBlocks: number,
+  // Input values
+  stakingAmountBTC: number,
```
vitsalis, line 12: We should always be using satoshis internally

src/utils/stakingSignReady.ts
```diff
@@ -0,0 +1,40 @@
...
+  finalityProvider: FinalityProvider | undefined,
+) => {
+  if (!finalityProvider) return false;
+
+  // API data is ready
+  const stakingAmountAPIReady = minStakingAmountSat && maxStakingAmountSat;
+  // App inputs are filled
+  const stakingAmountAppReady =
+    stakingAmountBTC >= satoshiToBtc(minStakingAmountSat) &&
+    stakingAmountBTC <= satoshiToBtc(maxStakingAmountSat);
```
vitsalis, line 23: All comparisons should be done with satoshis

src/utils/stakingSignReady.ts
```diff
@@ -0,0 +1,40 @@
+import { FinalityProvider } from "@/app/types/finalityProviders";
+import { satoshiToBtc } from "./btcConversions";
+
+// Check if the staking transaction is ready to be signed
+export const stakingSignReady = (
+  // API values
```
vitsalis, line 6: This component should have no knowledge on the existence or not of an API. I would rename to "parameter values" (same for the variable names)

src/utils/stakingSignReady.ts
```diff
@@ -0,0 +1,40 @@
...
+  const stakingAmountReady = stakingAmountAPIReady && stakingAmountAppReady;
+
+  // API data is ready
+  const stakingTimeAPIReady = minStakingTimeBlocks && maxStakingTimeBlocks;
+  // App inputs are filled
+  const stakingTimeAppReady =
+    stakingTimeBlocks >= minStakingTimeBlocks &&
+    stakingTimeBlocks <= maxStakingTimeBlocks;
+  // Staking time is fixed
+  const stakingTimeFixed = minStakingTimeBlocks === maxStakingTimeBlocks;
```
vitsalis, line 34: Why do we care if the staking time is fixed? If it is fixed, then `stakingTimeBlocks` should satisfy the previous condition.

src/app/components/Staking/Form/StakingAmount.tsx
```diff
@@ -0,0 +1,36 @@
+interface StakingAmountProps {
+  minStakingAmountBTC: number;
+  maxStakingAmountBTC: number;
+  stakingAmountBTC: number;
```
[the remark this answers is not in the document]
gbarkhatov: Done

src/app/components/Staking/Form/StakingAmount.tsx
```diff
@@ -0,0 +1,36 @@
...
+      </div>
+      <input
+        type="number"
+        placeholder="BTC"
+        className="no-focus input input-bordered w-full"
+        min={minStakingAmountBTC}
+        max={maxStakingAmountBTC}
+        step={0.00001}
+        value={stakingAmountBTC}
+        onChange={(e) => onStakingAmountChange(Number(e.target.value))}
```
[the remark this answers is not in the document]
gbarkhatov: Done

src/app/components/Modals/PreviewModal.tsx
```diff
@@ -95,10 +97,7 @@ export const PreviewModal: React.FC<PreviewModalProps> = ({
           </button>
           <button
             className="btn-primary btn flex-1"
-            onClick={() => {
-              onClose(false);
-              onSign();
-            }}
+            onClick={() => onSign(amountSat, stakingTimeBlocks)}
```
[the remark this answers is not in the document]
gbarkhatov: It doesn't know about that, it just passes the `amountSat` and `stakingTimeBlocks` to the parents' `handleSign`, which should receive these values. Without these values we are once again need to check whether or not we are in the fixed terms state or not, i.e. eventually go back to the root of the problem

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/app/components/Staking/Staking.tsx
gbarkhatov: > already in the state
minStakingTime comes from the API, it's not our react state. Fixed term passes this to the Modal, modal passes this to the Staking

src/app/components/Modals/PreviewModal.tsx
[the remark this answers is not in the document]
gbarkhatov: Done

src/app/components/Staking/Form/StakingAmount.tsx
```diff
@@ -0,0 +1,36 @@
...
+  maxStakingAmountBTC: number;
+  stakingAmountBTC: number;
+  onStakingAmountChange: (inputAmountBTC: number) => void;
+}
+
+export const StakingAmount: React.FC<StakingAmountProps> = ({
+  minStakingAmountBTC,
+  maxStakingAmountBTC,
+  stakingAmountBTC,
+  onStakingAmountChange,
```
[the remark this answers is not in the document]
gbarkhatov: Done

src/app/components/Staking/Form/StakingTime.tsx
```diff
@@ -0,0 +1,56 @@
+import { blocksToTime } from "@/utils/blocksToTime";
+
+interface StakingTimeProps {
+  minStakingTimeBlocks: number;
+  maxStakingTimeBlocks: number;
+  stakingTimeBlocks: number;
+  onStakingTimeChange: (inputTimeBlocks: number) => void;
```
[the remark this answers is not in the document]
gbarkhatov: Done

src/app/components/Staking/Staking.tsx
gbarkhatov: Formats on new line, adds |

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/app/components/Staking/Staking.tsx
gbarkhatov: It renders based on fps list, so fp will always be found. I can remove this `if`

src/app/components/Staking/Staking.tsx
gbarkhatov: If it's not set then theres going to be nothing to select from
I need

```ts
    if (!finalityProviders) {
      return;
    }
```
protection since Typescript tells us that fps are undefined at the time the component is rendered. But at the time this event is triggered fps are definitely defined since it can be triggered only when the data is there.

we can do this 
```ts
    const found = finalityProviders!.find((fp) => fp?.btcPk === btcPkHex);
    setFinalityProvider(found);
```
if you want, but I don't like exclamation marks and prefer more intuitive approach of `if data is not there - return` that we use everywhere. Easier to reason about

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/utils/stakingSignReady.ts
gbarkhatov: Done

src/utils/stakingSignReady.ts
gbarkhatov: Done

src/utils/stakingSignReady.ts
gbarkhatov: Done

src/utils/stakingSignReady.ts
gbarkhatov: If it's fixed then `stakingTimeBlocks` will be 0 (we're not setting this part of the state if we are using fixed term API params) and the condition won't be satisfied

src/app/components/Staking/Form/StakingTime.tsx
```diff
@@ -0,0 +1,56 @@
...
+  if (isFixed) {
+    return (
+      <div className="card mb-2 bg-base-200 p-4">
+        <p>
+          Your Signet BTC will be staked for a fixed term of{" "}
+          {blocksToTime(minStakingTimeBlocks)}.
+        </p>
+        <p>
+          You can unbond and withdraw your Signet BTC anytime through this
+          dashboard with an unbond time of 10 days.
```
vitsalis, line 26: It's gonna be 7 days.

src/app/components/Staking/Staking.tsx
vitsalis: But if this condition happens doesn't this mean that there's a programmer's error or something going wrong? I suggest we throw an error if something unexpected is happening. Let's be defensive: better to crash than something going wrong.

src/utils/isStakingSignReady.ts
```diff
@@ -0,0 +1,32 @@
...
+  const amountValuesReady = amount >= minAmount && amount <= maxAmount;
+  // Amount is ready
+  const amountIsReady = amountParamatersReady && amountValuesReady;
+
+  // Time parameters are ready
+  const timeParametersReady = minTime && maxTime;
+  // App values are filled
+  const timeValuesReady = time >= minTime && time <= maxTime;
+  // Staking time is fixed
+  const timeIsFixed = minTime === maxTime;
```
vitsalis, line 27: I don't like this condition happening here. Let's have the caller pass the correct staking time, either the fixed or the variable one. A util shouldn't handle this piece of logic.

vitsalis (approved):

src/app/components/Staking/Form/StakingTime.tsx
gbarkhatov: Changed to 7 days everywhere

src/utils/isStakingSignReady.ts
gbarkhatov: Done

PR #96, opened 2024-05-18
feature: Forward ref

### LINKED ISSUE #82: React modal not applying the theme properly using next-themes
CleanShot 2024-05-17 at 01 19 42@2x

Noticed the react-modal sometimes is not reactive to the main div theme switches. Resulting in white text and white background.

gbarkhatov:
Hard to reproduce issue, but the problem with the Modals losing the colors is they being rendered outside the `<main />` tag
Passing `page.tsx` ref to each of them should solve the issue

Should close #82


vitsalis:
Why is it difficult to reproduce? Don't completely understand the solution.

gbarkhatov:
> Why is it difficult to reproduce
Because I cannot reliably reproduce this on my env, let me close this PR as I think I found another solution

PR #97, opened 2024-05-18
feature: Theme

### LINKED ISSUE #82: React modal not applying the theme properly using next-themes

gbarkhatov:
Solves the issue with the theme. 
The problem was that our `tailwind` + `daisyui` + `next-themes` did not work properly using the latest updates. I've had to search for the solution using different unmerged PRs of these libraries.

- fix the modal colors
- remove .dark class
- changes of styles in small connect (top right corner)
- rm `useTheme` hook, using the `useTheme` from `next-themes` with `const { resolvedTheme } = useTheme();`
- rm 2 other unused hooks (`useMediaQuery` and `useDevice`) as we are using more reliable `css` approaches
- show a placeholder on `theme-toggle`. in case the theme is not resolved, there are not going to be any glitches related to the toggle showing usage of the light theme while the dark theme is actually used
- fixed background. changed the pattern (took only one rhombus from the svg), added proper scrolling solution so the background size doesn't "jump" when we're opening FAQ section and the height of the page changes. Aligned differently

Additionally, protected the Babylon logo:
The issue is that until the theme mounts, we don't know whether or not we are on a white or black theme. Later on, after all refactorings we are to check out the performance of the app since our *initial* load is somewhat heavy. While it's happening, the loaders are showing, but we don't know what version of Babylon logo should we use, since we are not using server-side rendering, nor static site generation. While this happens we have two versions of the Babylon logo - black or white. If we don't know which theme we're using, version mismatch will look very bad - the black logo would be barely seen on the dark header, the white logo would be invisible on the white background header. Added a react-content-loader with Babylon logo in primary color https://skeletonreact.com/

If it's considered overkill or we just want to show an empty logo - we can remove that part. Then the logo will be shown only after we know what theme we're using.

Closes #82


src/app/components/Header/Logo.tsx
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <ContentLoader
+        speed={1}
+        width={159}
+        height={40}
+        viewBox="0 0 159 40"
+        backgroundColor="#F3823A"
+        foregroundColor="#ecebeb"
+      >
+        <path
```
vitsalis, line 33: What is this?

vitsalis (approved):
Not proficient in these stuff, but lgtm

src/app/components/Header/Logo.tsx
gbarkhatov: You can open react skeleton link to see the loader. 

Can be removed (empty space instead of logo)

Or logo to be replaced with gray svg version (will be both visible on black and white theme)

gbarkhatov:
@vitsalis removed the skeleton loader

PR #135, opened 2024-05-20
add new message for current block height vs activation block height

jeremy-babylonchain:
CleanShot 2024-05-21 at 02 19 30

note: the block number in the above is using the correct param, since its testnet and the block has already been reached, Im just displaying it to show the format as example.

### DIFF
--- a/src/app/components/Staking/Staking.tsx
+++ b/src/app/components/Staking/Staking.tsx
@@ -69,11 +69,14 @@ export const Staking: React.FC<StakingProps> = ({
     useState<FinalityProviderInterface>();
   const [previewModalOpen, setPreviewModalOpen] = useState(false);
   const [currentBlockHeight, setCurrentBlockHeight] = useState(0);
+  const { showError } = useError();
   const pollingInterval = 60000; // 60 seconds fetch interval
 
   const stakingParams = paramWithContext?.currentVersion;
   const isUpgrading = paramWithContext?.isApprochingNextVersion;
-  const { showError } = useError();
+  const isBlockHeightUnderActivation =
+    !stakingParams || currentBlockHeight < stakingParams.activationHeight;
+  
 
   const handleResetState = () => {
     setFinalityProvider(undefined);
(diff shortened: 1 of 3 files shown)


jeremy-babylonchain:
@jrwbabylonchain

```jsx
// 3. Staking has not started yet
    else if (!stakingParams) {
      return (
        <Message
          title="Staking has not yet started"
          messages={[
            "The staking application will open once the staking activation height has been reached.",
          ]}
          icon={stakingNotStarted}
        />
      );
    }
```

Should we change the message displaying in this condition? since `activationHeight` is in `stakingParams`, I can't put the current height vs activation height message here.

src/app/components/Staking/Staking.tsx
```diff
@@ -71,6 +75,28 @@ export const Staking: React.FC<StakingProps> = ({
     setPreviewModalOpen(false);
   };
 
+  useEffect(() => {
+    let intervalId;
+
+    const fetchBlockHeight = async () => {
```
jrwbabylonchain, line 81: i think this data has already been fetch in the page.tsx with useQuery. and it will be fetched periodically.
It was fetched to calculate the global param. I think you can modify that method instead to avoid making multiple duplicated calls

src/app/components/Staking/Staking.tsx
```diff
@@ -198,7 +224,20 @@ export const Staking: React.FC<StakingProps> = ({
         />
       );
     }
-    // 6. Staking form
+    // 6. Current block height has not reached activation block height
+    else if (isBlockHeightUnderActivation) {
```
jrwbabylonchain, line 228: duplicated of https://github.com/babylonchain/simple-staking/pull/135/files#diff-5fedde225f959de8381d4e4314e6195dc7c15af4b34d63953c181ca2d74488e0R190
You should add the updated text there

jrwbabylonchain (approved):

gbarkhatov (approved):

PR #136, opened 2024-05-20
feature: Staking modal mobile + fixed term conversions

gbarkhatov:
1. Preview modal fix on mobile so it shows properly
- Before: https://i.imgur.com/BkcxF92.png
- Now: https://i.imgur.com/4ZJ4oxi.png
2. `blocksToTime` receives optional second argument to do the conversion. Solves our fixed term issue. Now instead of `1 year 2 months 20 days 10 hours 40 minutes` we can show `63 weeks`. If we later on want to show the days, minutes, years, whatever - it supports this


src/utils/blocksToTime.ts
```diff
@@ -19,26 +17,36 @@ const MINUTES_IN_HOUR = 60;
...
+export const blocksToTime = (
+  blocks: number,
+  unit?: keyof Duration,
+  precision?: number,
+): string => {
+  // If no blocks are provided, return "-"
   if (!blocks) return "-";
 
+  // If no unit is provided, return the number of blocks
+  if (!unit) return `${blocks} blocks`;
```
jrwbabylonchain, line 48: This seems confusing as the method name is about `time`. 

Why not default to a time that is calculated by something like this? https://momentjs.com/docs/#/displaying/fromnow/
i.e we don't need to manually do it.

jrwbabylonchain (approved):

src/utils/blocksToTime.ts
vitsalis: Also, why return the `blocks` suffix here? We don't do it anywhere else.

src/utils/blocksToTime.ts
```diff
@@ -10,27 +8,89 @@ interface Duration {
...
+ * blocksToTime(20000, "months"); // "4 months"
+ * blocksToTime(30000, "weeks"); // "29 weeks"
+ * blocksToTime(15000, "days"); // "104 days"
+ * blocksToTime(500, "hours"); // "83 hours"
+ * blocksToTime(250, "minutes"); // "2500 minutes"
+ * blocksToTime(100, "seconds"); // "60000 seconds"
+ * blocksToTime(10000, "invalidUnit"); // "-"
+ * blocksToTime(64000, "weeks", 5); // "65 weeks" instead of "63 weeks"
+ */
+export const blocksToTime = (
```
vitsalis, line 39: Why have this complex function when what we need is a method to convert blocks to weeks? Let's just have this, I don't see why we need to have functionality for supporting seconds/minutes/etc. and complicate the code.

If we want to have it, is there any existing functionality we could use? This function looks super complex.

vitsalis (changes requested):

src/utils/blocksToTime.ts
gbarkhatov: >This seems confusing as the method name is about time.
name is about time, but there's a unit of time

> Why not default to a time that is calculated by something like this
It was trickier to reason about, and then we have a requirement about rounding precision

>Also, why return the blocks suffix here? We don't do it anywhere else.
just return blocks string

src/utils/blocksToTime.ts
gbarkhatov: > Let's just have this, I don't see why we need to have functionality for supporting seconds/minutes/etc. and complicate the code.

We did not use "weeks only" before in our apps, why do we think now that the requirements won't change later on? In case of `non fixed` term for example, or after the testnet. I suggest keep it

vitsalis: Currently we are only using it to display a static unbonding time. Why make our life hard to have something super general that we don't need right now? When we need it we can introduce something that covers our new needs.

gbarkhatov: @vitsalis used date-fns and converted to the simple blocksToWeeks

src/utils/blocksToWeeks.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param {number} [precision] - The precision to round the result to. If not provided, the result is not rounded.
+ * @returns {string} - The converted time in weeks, rounded to the specified precision.
+ *
+ * Examples of usage:
+ * blocksToWeeks(30000); // "29 weeks"
+ * blocksToWeeks(63000, 5); // "65 weeks"
+ */
+export const blocksToWeeks = (blocks: number, precision?: number): string => {
+  // If no blocks are provided, return "-"
+  if (!blocks) return "-";
```
vitsalis, line 18: Let's throw an error here, it should not be called with an empty parameter.

vitsalis (approved):

PR #141, opened 2024-05-20
feature: Overflow state and tooltip is shown on active

gbarkhatov:
`isOverflow && status = "Active"` -> status and tooltip is overflow, otherwise show the real status

### DIFF
--- a/src/app/components/Delegations/Delegation.tsx
+++ b/src/app/components/Delegations/Delegation.tsx
@@ -73,8 +73,13 @@ export const Delegation: React.FC<DelegationProps> = ({
     }
   };
 
+  const isActive =
+    intermediateState === DelegationState.ACTIVE ||
+    state === DelegationState.ACTIVE;
+
   const renderState = () => {
-    if (isOverflow) {
+    // overflow should be shown only on active state
+    if (isOverflow && isActive) {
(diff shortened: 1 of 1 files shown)


src/app/components/Delegations/Delegation.tsx
```diff
@@ -73,16 +73,22 @@ export const Delegation: React.FC<DelegationProps> = ({
     }
   };
 
+  const isActive =
+    intermediateState === DelegationState.ACTIVE ||
```
jrwbabylonchain, line 77: question regarding this `intermediateState` props. why is this being passed in from parent instead being computed as a react state inside this component? If i understood correctly, this is the state after user take some actions on the delegation component, such as unbonding etc.

jrwbabylonchain (approved):

vitsalis (approved):

src/app/components/Delegations/Delegation.tsx
gbarkhatov: True, it is a local storage information that we do outside. The split was done before that when we were discussing things with @vitsalis about the splitting the logic with API state and local storage state

Anyways - this ticket is about the small UI fix, let's create an architectural ticket (if needed) and solve it there

PR #144, opened 2024-05-21
feature: Add OneKey Wallet Provider

originalix:
Add the [OneKey](https://onekey.so) wallet entry to support the wallet selector


src/utils/wallet/onekey_wallet.ts
```diff
@@ -0,0 +1,61 @@
+import { Fees, Network, UTXO, WalletProvider } from "./wallet_provider";
+
+export class OneKeyWallet extends WalletProvider {
+  private provider = window.$onekey.btcwallet;
+
+  async connectWallet(): Promise<this> {
+    return this.provider.connectWallet();
```
gbarkhatov, line 7: it should return an address and public key
```
      this.oneKeyWalletInfo = {
        publicKeyHex: compressedPublicKey,
        address,
      };
      return this;
```

src/utils/wallet/onekey_wallet.ts
```diff
@@ -0,0 +1,61 @@
+import { Fees, Network, UTXO, WalletProvider } from "./wallet_provider";
+
+export class OneKeyWallet extends WalletProvider {
+  private provider = window.$onekey.btcwallet;
```
gbarkhatov, line 4: nice approach
but one concern - onekey is currently occupying window.btcwallet, which will interfere with another wallets

src/utils/wallet/onekey_wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  async getNetwork(): Promise<Network> {
+    return this.provider.getNetwork();
+  }
+
+  async signMessageBIP322(message: string): Promise<string> {
+    return this.provider.signMessageBIP322(message);
+  }
+
+  on(eventName: string, callBack: () => void): void {
```
gbarkhatov, line 38: does this work? "accountChanged" event

src/utils/wallet/onekey_wallet.ts
```diff
@@ -0,0 +1,61 @@
+import { Fees, Network, UTXO, WalletProvider } from "./wallet_provider";
```
gbarkhatov, line 1: connecting onekey will return `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 debugReady listeners added` warning

src/utils/wallet/onekey_wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  async getAddress(): Promise<string> {
+    return this.provider.getAddress();
+  }
+
+  async getPublicKeyHex(): Promise<string> {
+    return this.provider.getPublicKeyHex();
+  }
+
+  async signPsbt(psbtHex: string): Promise<string> {
```
gbarkhatov, line 22: `signPsbt` should return finalized PSBT hex.
- Signing staking transaction will return `Format Error: Invalid Magic Number` unless I use `autoFinalized: true` option: `return window.$onekey?.btcwallet.signPsbt(psbtHex, { autoFinalized: true });`
- Signing unbonding transaction will not work with the above approach. We should use instead:
1. using it without `autoFinalized`. It will return Transaction hex.
2. convert from the Transaction hex.

But we expect the return to be finalized PSBT hex string

src/utils/wallet/onekey_wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  async signMessageBIP322(message: string): Promise<string> {
+    return this.provider.signMessageBIP322(message);
+  }
+
+  on(eventName: string, callBack: () => void): void {
+    this.provider.on(eventName, callBack);
+  }
+
+  async getBalance(): Promise<number> {
```
gbarkhatov, line 42: returns zero string: `'0'` 
if it's not implemented we should fallback to mempool

src/utils/wallet/onekey_wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  async getNetworkFees(): Promise<Fees> {
+    return this.provider.getNetworkFees();
+  }
+
+  async pushTx(txHex: string): Promise<string> {
+    return this.provider.pushTx(txHex);
+  }
+
+  async getUtxos(address: string, amount: number): Promise<UTXO[]> {
```
gbarkhatov, line 54: returns an empty array, should use the mempool fallback if not implemented

gbarkhatov (changes requested):

src/utils/wallet/onekey_wallet.ts
originalix: `this.oneKeyWalletInfo` or `this.walletInfo` is not defined in the Provider. Will you try to access this member variable?

If it needs to be accessed, it would be best to define it within the class.

src/utils/wallet/onekey_wallet.ts
originalix: Because the global variable provided in the documentation is `window.btcwallet`, we will register it and provide the variables under the `$onekey` property.

Similar to how `window.ethereum` and `window.unisat` are handled.

However, it's okay— in the next version, we will update a feature to allow users not to set OneKey as the default injected wallet. For the current version, it can only remain this way.

src/utils/wallet/onekey_wallet.ts
originalix: `accountsChanged` is working correctly.

https://docs.unisat.io/dev/unisat-developer-center/unisat-wallet#accountschanged

Got it, I didn't notice that Babylon uses `accountChanged`. If necessary, we will fix this issue in the next version.

src/utils/wallet/onekey_wallet.ts
originalix: Thank you for the reminder. We have also noticed this issue and will fix it as soon as possible. 🙏

src/utils/wallet/onekey_wallet.ts
originalix: CleanShot 2024-05-22 at 19 52 50@2x

getBalance is working correctly. Please confirm that your plugin version is 4.24.0.

originalix: However, I feel that using mempool is a more reliable solution, and it also makes it easier to unify the implementations across different providers.

Fixed.

src/utils/wallet/onekey_wallet.ts
originalix: Fixed

originalix: It should be noted that if you use mempool for querying, you can only retrieve the balance and UTXO of the current address. The balance and UTXO of change addresses and newly derived addresses cannot be queried.

src/utils/wallet/onekey_wallet.ts
originalix: https://gist.github.com/vitsalis/7b007abc82b795c49460f994d64364e9#3-example-okx-wallet

```ts
// sign the PSBTs
const response =
  await window?.okxwallet?.bitcoinSignet?.signPsbts(psbtsHexes);

// convert the signed PSBTs to transactions
return response.map((tx: any) =>
  Psbt.fromHex(tx).extractTransaction().toHex(),
);
```

OneKey's implementation references the original code snippet provided by OKX. We can see that the previous signPsbt method processes the psbtHex further by doing `Psbt.fromHex(tx).extractTransaction().toHex()`

Thus, we performed the same processing; otherwise, signing would fail on older versions of the dApp.

originalix: Currently, our app implementation has already been launched. Could you make it compatible at the Provider layer?

Another version update won't make it in time for the testnet.

originalix: The specific implementation is here 

https://github.com/OneKeyHQ/app-monorepo/blob/56b2c03a732e31837b8e4e64b3c66e05b92702ee/packages/kit-bg/src/providers/ProviderApiBtc.ts#L606

originalix: 😂

src/utils/wallet/onekey_wallet.ts
gbarkhatov: Any estimates?

src/utils/wallet/onekey_wallet.ts
gbarkhatov: Besides issues and conflicsts with other wallets there's another concern.

When we click "connect" we fallback to `window.btcwallet` to support the mobile injection mechanism. If your implementation in `onekey_wallet` is different from what you push inside `window.btcwallet` it will cause issues.

src/utils/wallet/onekey_wallet.ts
gbarkhatov: You can check out whenever we use this (to get an address or pk hex)

Also we have a new version on `main` branch with a list of wallets, you can rebase

src/utils/wallet/onekey_wallet.ts
gbarkhatov: @originalix So what is your suggestion?

src/utils/wallet/onekey_wallet.ts
gbarkhatov: Can you support both?

src/utils/wallet/onekey_wallet.ts
gbarkhatov: Without the mempool, the wallet returns a string, but expect to return a number

src/utils/wallet/onekey_wallet.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.provider.on(eventName, callBack);
+  }
+
+  async getBalance(): Promise<number> {
+    // mempool call
+    return getAddressBalance(await this.getAddress());
+  }
+
+  async getNetworkFees(): Promise<Fees> {
+    return this.provider.getNetworkFees();
```
gbarkhatov, line 52: What about the other fields? Are they compatible with the return types? i.e. I can swap mempool implementation with Onekey implementation and receive the same results?

gbarkhatov:
@originalix I suggest you rebasing with the main branch, adding onekey to the list and checking out the full cycle - connecting, staking, (waiting a bit), unbonding

originalix:
> @originalix I suggest you rebasing with the main branch, adding onekey to the list and checking out the full cycle - connecting, staking, (waiting a bit), unbonding

```
if (window.btcwallet) {
      handleConnectBTC(window.btcwallet);
    } else {
      // show a wallet selection modal
      setConnectModalOpen(true);
    }
```

This logic is quite strange. In the initial integration documentation, it was required for wallets to implement the `window.btcwallet` object. According to this logic, wallets that have implemented the `window.btcwallet` object will not trigger the wallet selector.

src/utils/wallet/onekey_wallet.ts
originalix: The return type of getNetworkFees is also a string, but the expected type is a number. For now, we are using a mempool fallback.

This is expected to be fixed in the next version.

originalix:
All issues have been fixed. Please use the wallet selector to connect to the OneKey wallet, and you should be able to stake successfully.

gbarkhatov:
@originalix 

1. Add your wallet to list.ts. Please use your icon and add a link docs. There's a provider field that is a constant, extracted from your `onekey_wallet.ts`. 
```ts
{
  name: "OneKey",
  icon: oneKeyIcon,
  wallet: OneKeyWallet,
  provider: oneKeyProvider,
  linkToDocs: "https://onekey.so/download",
},
```
2. Implement `onekey_wallet.ts`. Create a window constant:
```ts
export const oneKeyProvider = "$onekey";
```

and use it like 
```ts
window[oneKeyProvider]?.btcwallet
```
3. Add your wallet window provider to the window `src/utils/types.ts`
```ts
declare global {
  interface Window {
    btcwallet: WalletProvider;
    btc: any;
    keplr: any;
    okxwallet: any;
    $onekey: any;
  }
}
```
4. Try to do a full cycle: `connect -> stake -> unbond`.

Your current implementation has issues with unbonding. What you do with `window.btcwallet` does not work.

Please rebase once this is merged https://github.com/babylonchain/simple-staking/pull/164

originalix:
After taking a closer look at the process, I noticed that the implementation differences are due to the `signPsbts` method in the OKX example within `simple-staking/docs/WalletIntegration.md`. In your implementation, you placed `Psbt.fromHex(tx).extractTransaction()` inside the `OKXProvider`. Consequently, I implemented it inside our App, which is why OneKey always returns a Transaction hex string.

My suggested solution is to add a `convertSignedPsbtToTransaction` method to the Provider. This would eliminate the implementation differences across various Providers. I have already tested this solution, and the `connect -> stake -> unbond` process works successfully with both OKX and OneKey wallets.

When OneKey release the new version, I will submit a PR to fix these issues. At that time, we can evaluate together whether we should remove the `convertSignedPsbtToTransaction` method.

originalix:
I noticed that your wallet selector automatically detects if a wallet is installed, which is a great implementation.

I think we can maintain this on mobile as well by displaying the wallet selector. This way, mobile users can also connect using `OneKeyProvider` or `OKXProvider`.

vitsalis:
Hi @originalix, I don't think adding the `convertSignedPsbtToTransaction` is a good idea as it complicates the interface and creates inconsistencies as the dApp does not know what to expect from `signPsbt`. Further, our wallet collaborators are already using the current format, so it is difficult to change the existing interface prior to the testnet launch.

The outdated example about OKX on the Wallet Integration guide has been updated in #165 . How about we do the following: Given that it is difficult to update your wallet extension at the moment to have a provider that satisfies the interface, why don't you perform the necessary conversions that are needed to return the expected format inside your `signPsbt` transaction in this PR?

originalix:
Since our version has already been released, it will take some time to release another version. The next release is expected by the end of June.

The changes from the OKX PR were not synced to the documentation. The `simple-staking` I used for testing before the release has not yet merged #165.

Therefore, I suggest using `convertSignedPsbtToTransaction` as a temporary solution. Once the new version of OneKey fixes the issue, we can remove this interface.

If you have any better suggestions, feel free to propose them or make changes directly in the code. I just provided an example to ensure the every wallet works correctly.

jrwbabylonchain:
fyi, i opened a PR against your branch to refectories the code a bit: https://github.com/originalix/simple-staking/pull/1

gbarkhatov (approved):

jrwbabylonchain (approved):

PR #145, opened 2024-05-21
feat(wallet): add the Bitget Wallet

DetonateYin2077:
Hi babylonchain team,

This PR is to add the Bitget Wallet interface that wraps the methods exposed followed by [WalletIntegration](https://github.com/babylonchain/simple-staking/blob/main/docs/WalletIntegration.md)
Please let me know if you have any concerns.

**Type of Change**:
> New Feature

**Checklist**:

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have linted the codes and got no ESLint warnings or errors

Bitget Wallet team


vitsalis:
Hi @DetonateYin2077 , it seems that this code invokes `window.btcwallet` which is the attribute that should be filled with an object of class you create. Looking at [this](https://web3.bitget.com/en/docs/provider-api/btc.html), isn't the Bitget wallet API the `window.bitget.unisat`?

DetonateYin2077:
> Hi @DetonateYin2077 , it seems that this code invokes `window.btcwallet` which is the attribute that should be filled with an object of class you create. Looking at [this](https://web3.bitget.com/en/docs/provider-api/btc.html), isn't the Bitget wallet API the `window.bitget.unisat`?

Hi @vitsalis , yes, our official API is window.bitget.unisat. However, during our previous communication for the bbn-test-3 phase, based on [WalletIntegration](https://gist.github.com/vitsalis/7b007abc82b795c49460f994d64364e9), we provided a unified API named window.btcwallet for both PC and mobile platforms specifically for Babylon (refer to [this](https://gist.github.com/vitsalis/7b007abc82b795c49460f994d64364e9)). Therefore, to ensure smooth support for the event, we plan to continue using this API in the current PR. We will update the wallet adapter according to the 'window.bitget.unisat' API in the future.

vitsalis:
@DetonateYin2077 the `window.btcwallet` interface is the attribute that the dApp uses internally after the wallet has been connected. More specifically, once the user selects the wallet that they want to connect, the app sets `window.btcwallet` to that wallet's interface. We have designed it this way to allow mobile app wallets to inject their wallet into the dApp when opened by the in-app browser if the wallet was not already supported by the app. 

However, the Bitget browser extension exposing `window.btcwallet` directly leads to issues in the application. More specifically, the application sees the `window.btcwallet` attribute as set and thinks that a wallet is already connecting, therefore not allowing users to select the wallet they want to connect.

Due to the above, the Bitget browser extension should update to not set the `window.btcwallet` attribute.  For example, OneKey wallet sets the `window.$onekey.btcwallet` attribute ([RP](https://github.com/babylonchain/simple-staking/pull/144)). Bitget can do something similar.

> We will update the wallet adapter according to the window.bitget.unisat API in the future.
Good point on making this update in the future.

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,69 @@
...
+  signPsbts = async (psbtsHexes: string[]): Promise<string[]> => {
+    return window?.btcwallet?.signPsbts(psbtsHexes);
+  };
+
+  signMessageBIP322 = async (message: string): Promise<string> => {
+    return window?.btcwallet?.signMessageBIP322(message);
+  };
+
+  getNetwork = async (): Promise<Network> => {
+    return window?.btcwallet?.getNetwork();
```
gbarkhatov, line 43: return `undefined`

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  on = (eventName: string, callBack: () => void) => {
+    return window?.btcwallet?.on(eventName, callBack);
+  };
+
+  getBalance = async (): Promise<number> => {
+    return window?.btcwallet?.getBalance();
```
gbarkhatov, line 51: throws an error `{code: -32603}`

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  getNetworkFees = async (): Promise<Fees> => {
+    return window?.btcwallet?.getNetworkFees();
```
gbarkhatov, line 55: throws an error `{code: -32603}`

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  pushTx = async (txHex: string): Promise<string> => {
```
gbarkhatov, line 58: these methods should rely to mempool if not implemented on the wallet side

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  signMessageBIP322 = async (message: string): Promise<string> => {
+    return window?.btcwallet?.signMessageBIP322(message);
+  };
+
+  getNetwork = async (): Promise<Network> => {
+    return window?.btcwallet?.getNetwork();
+  };
+
+  on = (eventName: string, callBack: () => void) => {
```
gbarkhatov, line 46: does this work?

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  getAddress = async (): Promise<string> => {
+    return window?.btcwallet?.getAddress();
+  };
+
+  getPublicKeyHex = async (): Promise<string> => {
+    return window?.btcwallet?.getPublicKeyHex();
+  };
+
+  signPsbt = async (psbtHex: string): Promise<string> => {
```
gbarkhatov, line 30: `signPsbt` should return finalized PSBT hex.
At the moment - signing unsigned staking PSBT works, signing unsigned unbonding does not (tested on Taproot address). It returns the same `unsignedUnbondingTx` that was passed to the `signPsbt`.
No errors on UI side of the modal extension, when you click "agree" it just returns the same string

gbarkhatov (changes requested):

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,69 @@
+import { WalletProvider, Network, Fees, UTXO } from "./wallet_provider";
+import { getTipHeight } from "../mempool_api";
+
+export class BitgetWallet extends WalletProvider {
+  constructor() {
```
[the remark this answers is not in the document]
DetonateYin2077: Done

> it should have a private `bitgetWalletInfo` field with `publicKeyHex: string; address: string;`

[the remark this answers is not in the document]
DetonateYin2077: > Additionally, it can have `private provider = window.bitkeep?.unisat;` or whatever you use to call the exposed methods. Or you should manually call `window.wallet.someMethod` below

Good point. Just added it.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Could u be more spercificly pls?

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Nailed It.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Checkout the rewrited version pls

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Checkout the rewrited version pls

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Checkout the rewrited version pls

src/utils/wallet/bitget_wallet.ts
gbarkhatov: We use `accountChanged` event at the moment. In case the wallet is connected and then the user switches from Taproot to Native SegWit the app will automatically connect to wallet once again, receiving it's address, balance, pk hex, etc. If I try to pass the same stuff using BitGet it doesn't work as the current (OKX) implementation

src/utils/wallet/bitget_wallet.ts
gbarkhatov: @DetonateYin2077 did you try to unbond? It shows no error on the wallet ui side, but it returns the same string that was passed

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,174 @@
...
+        }
+      }
+    }
+
+    let signedPsbt = await this.provider?.request('dappsSign', data)
+    let psbt = Psbt.fromHex(signedPsbt)
+    try {
+      psbt.finalizeAllInputs()
+    } catch (error) {
+      console.log('babylon-signPsbt failed', error)
```
vitsalis, line 90: This should throw an error

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  signMessageBIP322 = async (message: string): Promise<string> => {
+    return await this.provider?.signMessage(message, 'bip322-simple')
+  };
+
+  getNetwork = async (): Promise<Network> => {
+    let network = await this.provider?.getNetwork()
+
+    if (network === 'signet') {
+      return 'testnet'
```
vitsalis, line 144: Why does this return `testnet` here instead of `signet`?

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,174 @@
...
+  getNetworkFees = async (): Promise<Fees> => {
+    return await getNetworkFees();
+  };
+
+  pushTx = async (txHex: string): Promise<string> => {
+    return await pushTx(txHex);
+  };
+
+  getUtxos = async (address: string, amount: number): Promise<UTXO[]> => {
+    // mempool call
```
vitsalis, line 167: Comment covered by the prior one.

src/utils/wallet/index.ts
```diff
@@ -7,7 +8,8 @@ const taprootAddressLength = 62;
 
 // Get the wallet provider from the window object, default to OKXWallet if not found.
 export const getWallet = (): WalletProvider => {
-  return window.btcwallet ? window.btcwallet : new OKXWallet();
+  return new BitgetWallet()
+  // return window.btcwallet ? window.btcwallet : new BitgetWallet();
```
vitsalis, line 12: We can handle this particular update for wallet selection, no need to do this change in this PR.

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    if (address && publicKeyHex) {
+      this.bitgetWalletInfo = {
+        publicKeyHex,
+        address,
+      };
+      return this;
+    } else {
+      throw new Error("Could not connect to Bitget Wallet");
+    }
```
vitsalis, line 51:
```suggestion
    if (!address || !publicKeyHex) {
      throw new Error("Could not connect to Bitget Wallet");
    }
    this.bitgetWalletInfo = {
      publicKeyHex,
      address,
    };
    return this;
```

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,174 @@
...
+  };
+
+  getWalletProviderName = async (): Promise<string> => {
+    return 'Bitget Wallet';
+  };
+
+  getAddress = async (): Promise<string> => {
+    let accounts = (await this.provider?.getAccounts()) || []
+    return accounts[0]
```
vitsalis, line 60: This will crash the app if the `accounts == []`.

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,174 @@
...
+  };
+
+  getPublicKeyHex = async (): Promise<string> => {
+    let publicKey = await this.provider?.getPublicKey()
+    if (!publicKey) {
+      console.log('bitget Wallet not connected')
```
vitsalis, line 66: We should throw an error here

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+    return publicKey
+  };
+
+  signPsbt = async (psbtHex: string): Promise<string> => {
+    const data = {
+      method: 'signPsbt',
+      params: {
+        from: this.provider?.selectedAddress,
+        __internalFunc: '__signPsbt_babylon',
```
vitsalis, line 77: What is this `__internalFunc`?

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    try {
+      let signedPsbts = await this.provider?.request('dappsSign', data)
+      signedPsbts = signedPsbts.split(',')
+      return signedPsbts.map((tx:string) => {
+        let psbt = Psbt.fromHex(tx)
+        try {
+          psbt.finalizeAllInputs()
+        } catch (error) {
+          console.log('babylon-signPsbts failed', error)
```
vitsalis, line 125: this should throw an error.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Aligned with the code at https://github.com/babylonchain/simple-staking/blob/c888a65cc339be83aa13fac3fc70a5811543ecc9/src/utils/wallet/index.ts#L18

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: @gbarkhatov I just finish the staking tests and got not transaction to be unbounded. Could you provide the inputs for unbundling on TG so I can test this procedure?

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: The key point is to ensure that all PSBTs are finalized, and we do not want it to throw an error if the PSBT has already been finalized. Therefore, if this procedure is not needed, the try-catch block could be removed, but there is no guarantee that the PSBT will be finalized.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Same reason with the signPsbt. Should we remove it or leave it as is?

src/utils/wallet/bitget_wallet.ts
vitsalis: If `finalizeAllInputs()` throws an error won't this mean that the psbt will eitherway not be finalized? If we care about the case where the psbt has already been finalized can we check this some other way instead of catching an exception here?

DetonateYin2077: We can achieve this in other ways, but there are scenarios in which the finalize process might fail in practice. For example, in a 2-of-3 multisig transaction, a PSBT might fail to finalize if only one signature has been collected. In such cases, the partially signed PSBT can be sent to the next signer for further processing even if finallized failed. We use try-catch to accommodate these scenarios.

Does the babylon Dapp handle such cases? Or can we ensure that the PSBT here must be successfully finalized here, and if not, an exception should be thrown?

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Done

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Fixed

src/utils/wallet/index.ts
DetonateYin2077: Fixed.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Done.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Done

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Done

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,167 @@
+import { WalletProvider, Network, Fees, UTXO } from "./wallet_provider";
+import {
+  getAddressBalance,
+  getTipHeight,
+  getFundingUTXOs,
+  getNetworkFees,
+  pushTx,
+} from "../mempool_api";
+import { Psbt } from 'bitcoinjs-lib'
```
gbarkhatov, line 9: please put global imports above, then a separate line, then local imports

src/utils/wallet/bitget_wallet.ts
gbarkhatov: @DetonateYin2077 did you see any errors in the browser console? 
If staking went properly, you will see the unbonding button after some time

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,167 @@
...
+    try {
+      psbt.finalizeAllInputs()
+    } catch (error) {
+      console.log('babylon-signPsbt failed', error)
+    }
+
+    return psbt.toHex()
+  };
+
+  signPsbts = async (psbtsHexes: string[]): Promise<string[]> => {
```
gbarkhatov, line 97: If these methods work the same - can we reuse the logic so there's no duplication? I.e. signing 1 psbt is the same as signing psbts array with only 1 item

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (!psbtsHexes && !Array.isArray(psbtsHexes)) {
+      throw 'parmas error'
```
gbarkhatov, line 99: typo

gbarkhatov (changes requested):

src/utils/wallet/bitget_wallet.ts
vitsalis: Hmm I see. Let's expect the transaction to be successfully finalized here but not your concern in a comment. If the transaction is not finalized here then the rest of the application will break as it expects that the transaction is fully signed after this step.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: It is not the same for Bitget Wallet. One small difference is in the parameters: the `psbtHexs` property should be passed when using signPsbts instead of `psbtHex`.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Got it. I will change the logic here to ensure every PSBT is fully finalized, and if not, an error will be thrown.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Pardon?

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: @gbarkhatov After digging into this, here are my test result:
1. The input which need to be signed is 
70736274ff01005e0200000001c955cf89c972b04437dee177ea0edda95d1a46f4d723f1bf8b161d8f881b1da50000000000ffffffff01581b000000000000225120d6f992871480a7e00c40dfbfa9959c42e41c4b3ced27c54d5d27a2c1fdfb0cfd000000000001012b1027000000000000225120a3c508bccdb7b56c7676d83384d66329aedba609c929e2f51dc521b3c8638ac86215c150929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0ce390e5b1389eb57748833b52cc6d7db09dc72f6fef551f85bf4e08904ea4d149e1ef557d205f0a40672403150502c57b2144a63449b109f181f734ee2cf2065fd57012045e3ad0ce099aedac4658d4c68e772d90276c777a1fb3282c2e9f5ed9802225fad20113c3a32a9d320b72190a04a020a0db3976ef36972673258e9a38a364f3dc3b0ac2017921cf156ccb4e73d428f996ed11b245313e37e27c978ac4d2cc21eca4672e4ba203bb93dfc8b61887d771f3630e9a63e97cbafcfcc78556a474df83a31a0ef899cba2040afaf47c4ffa56de86410d8e47baa2bb6f04b604f4ea24323737ddc3fe092dfba2049766ccd9e3cd94343e2040474a77fb37cdfd30530d05f9f1e96ae1e2102c86eba2076d1ae01f8fb6bf30108731c884cddcf57ef6eef2d9d9559e130894e0e40c62cba2079a71ffd71c503ef2e2f91bccfc8fcda7946f4653cef0d9f3dde20795ef3b9f0ba20d21faf78c6751a0d38e6bd8028b907ff07e9a869a43fc837d6b3f8dff6119a36ba20f5199efae3f28bb82476163a7e458c7ad445d9bffb0682d10d3bdb2cb41f8e8eba559cc001172050929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac00000

2. This is the signed psbt hex:
70736274ff01005e0200000001c955cf89c972b04437dee177ea0edda95d1a46f4d723f1bf8b161d8f881b1da50000000000ffffffff01581b000000000000225120d6f992871480a7e00c40dfbfa9959c42e41c4b3ced27c54d5d27a2c1fdfb0cfd000000000001012b1027000000000000225120a3c508bccdb7b56c7676d83384d66329aedba609c929e2f51dc521b3c8638ac80108fdfd0103400d34227cfa8c879e202ec44d055f8a3edf783ddb5e9429ffe52018675cbfe72e8a980750d45a592cc20c680d8600c2ca7fbfc61f65b641b6042b9101eac0459efd56012045e3ad0ce099aedac4658d4c68e772d90276c777a1fb3282c2e9f5ed9802225fad20113c3a32a9d320b72190a04a020a0db3976ef36972673258e9a38a364f3dc3b0ac2017921cf156ccb4e73d428f996ed11b245313e37e27c978ac4d2cc21eca4672e4ba203bb93dfc8b61887d771f3630e9a63e97cbafcfcc78556a474df83a31a0ef899cba2040afaf47c4ffa56de86410d8e47baa2bb6f04b604f4ea24323737ddc3fe092dfba2049766ccd9e3cd94343e2040474a77fb37cdfd30530d05f9f1e96ae1e2102c86eba2076d1ae01f8fb6bf30108731c884cddcf57ef6eef2d9d9559e130894e0e40c62cba2079a71ffd71c503ef2e2f91bccfc8fcda7946f4653cef0d9f3dde20795ef3b9f0ba20d21faf78c6751a0d38e6bd8028b907ff07e9a869a43fc837d6b3f8dff6119a36ba20f5199efae3f28bb82476163a7e458c7ad445d9bffb0682d10d3bdb2cb41f8e8eba559c61c150929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0ce390e5b1389eb57748833b52cc6d7db09dc72f6fef551f85bf4e08904ea4d149e1ef557d205f0a40672403150502c57b2144a63449b109f181f734ee2cf20650000

And these hex strings are not the same. I am not sure what exactly the problem you encountered was?

src/utils/wallet/bitget_wallet.ts
vitsalis: It should be `params error`

src/utils/wallet/bitget_wallet.ts
gbarkhatov: > The unbound request has been successfully called
Okay

> I am not sure what exactly the problem you encountered was?
When I tested this out, I just used `signPsbt` without ` __internalFunc` or things like `let signedPsbt = await this.provider?.request('dappsSign', data)` as there's no docs about this stuff

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Done.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Done.

src/utils/wallet/bitget_wallet.ts
```diff
@@ -0,0 +1,167 @@
...
+  };
+
+  getWalletProviderName = async (): Promise<string> => {
+    return 'Bitget Wallet';
+  };
+
+  getAddress = async (): Promise<string> => {
+    let accounts = (await this.provider?.getAccounts()) || []
+    if (!accounts?.[0]) {
+      throw new Error("bitget Wallet not connected");
```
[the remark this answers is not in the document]
DetonateYin2077: Done. Good catch.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Done.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Done.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Unfortunately, after investigation, we found that this application scenario is a function not covered by Bitget Wallet before, we will supplement this function in the latest release, and ensure that the test result. And the calling method in this PR will not be affected.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: Sorry for causing that trouble. Since these functions are specially prepared for Babylon, there are no documents about them yet. We will refine this part and either add additional documentation about these scenarios or supply them officially. We will update you guys ASAP.

src/utils/wallet/bitget_wallet.ts
DetonateYin2077: @gbarkhatov And we're using the `accountsChanged` event name to notify this event, which aligns with Unisat and some other wallets. So, I propose changing the code to listen for this event in Bitget Wallet adaptor.I’ve tested it in our pre-release version of the Chrome extension wallet, and it works when the address is changed.

And sorry we didn't mention that in our official document, we are eagerly working on a new version and will fix this part.

gbarkhatov:
@DetonateYin2077 we updated our code supporting multiple wallets connection.
you need to
1. add your type to `src/utils/types.ts`
2. implement wallet (you already did)
3. add a logo and wallet name to the `src/utils/wallet/list.ts`

gbarkhatov:
@DetonateYin2077 additionally - we rely on `window.btcwallet` field. If this field is present (for example on mobile when we're injecting) - use that, without a wallet select.

If BitGet wallet is installed and putting things to `window.btcwallet` we will see an error - https://i.imgur.com/1DWt03X.png Cannot fetch the global params

Please check out what you're putting inside `window.btcwallet`

gbarkhatov:
@DetonateYin2077 

1. Add your wallet to list.ts. Please use your icon and add a link docs. There's a provider field that is a constant, extracted from your `bitget_wallet.ts`. 
```ts
{
  name: "Bitget",
  icon: bitGetIcon,
  wallet: BitgetWallet,
  provider: bitGetProvider,
  linkToDocs: "https://web3.bitget.com",
},
```
2. Implement `bitget_wallet.ts`. Create a window constant:
```ts
export const bitGetProvider = "bitkeep";
```
and use it like 
```ts
window[bitGetProvider]?.unisat?.requestAccounts();
```
3. Add your wallet window provider to the window `src/utils/types.ts`
```ts
declare global {
  interface Window {
    btcwallet: WalletProvider;
    btc: any;
    keplr: any;
    okxwallet: any;
    bitkeep: any;
  }
}
```
4. Try to do a full cycle: `connect -> stake -> unbond`.

Please check out https://github.com/babylonchain/simple-staking/pull/164

src/utils/wallet/providers/bitget_wallet.ts
```diff
@@ -0,0 +1,175 @@
...
+      signedPsbts = signedPsbts.split(',')
+      return signedPsbts.map((tx:string) => {
+        let psbt = Psbt.fromHex(tx)
+
+        try {
+          const allFinalized = psbt.data.inputs.every(input => input.finalScriptWitness || input.finalScriptSig);
+          if(!allFinalized){
+            psbt.finalizeAllInputs()
+          }
+        } catch (error) {
```
vitsalis, line 129: Do we need this `catch` given the outer `catch`?

src/utils/wallet/providers/bitget_wallet.ts
```diff
@@ -0,0 +1,175 @@
...
+          throw new Error((error as Error)?.message)
+        }
+
+        return psbt.toHex()
+      })
+    } catch (error) {
```
vitsalis, line 135: What is the usefulness of this catch as it just rethrows the error?

vitsalis (commented):
Minor from my side. Let's wait for @gbarkhatov to test

DetonateYin2077:
> @DetonateYin2077
> 
> 1. Add your wallet to list.ts. Please use your icon and add a link docs. There's a provider field that is a constant, extracted from your `bitget_wallet.ts`.
> 
> ```ts
> {
>   name: "Bitget",
>   icon: bitGetIcon,
>   wallet: BitgetWallet,
>   provider: bitGetProvider,
>   linkToDocs: "https://web3.bitget.com",
> },
> ```
> 
> 2. Implement `bitget_wallet.ts`. Create a window constant:
> 
> ```ts
> export const bitGetProvider = "bitkeep";
> ```
> 
> and use it like
> 
> ```ts
> window[bitGetProvider]?.unisat?.requestAccounts();
> ```
> 
> 3. Add your wallet window provider to the window `src/utils/types.ts`
> 
> ```ts
> declare global {
>   interface Window {
>     btcwallet: WalletProvider;
>     btc: any;
>     keplr: any;
>     okxwallet: any;
>     bitkeep: any;
>   }
> }
> ```
> 
> 4. Try to do a full cycle: `connect -> stake -> unbond`.
> 
> Your current implementation works. What you do with `window.btcwallet` does not, hence we can use the version from the PR.
> 
> Please rebase once this is merged #164

Hi @gbarkhatov, I've already made those adjustments. Please check our latest code change. The full cycle: `connect -> stake -> unbond` has also been tested, and it works well.

src/utils/wallet/providers/bitget_wallet.ts
DetonateYin2077: Make sense,  I just removed it.

gbarkhatov (approved):

gbarkhatov (changes requested):
One thing I found while testing: 
Until wallet is unlocked nothing will be thrown.
@DetonateYin2077 Can you check the `connectWallet` method once again? Do you have the "connect" method or similar that unlocks the wallet before requesting the account and switching to signet?
```ts
await this.provider?.requestAccounts() // Connect to Bitget Wallet extension
await this.provider?.switchNetwork('signet')
```
Since it's not done automatically by calling this methods https://youtu.be/mzvkhGc7p88

DetonateYin2077:
> https://youtu.be/mzvkhGc7p88



> One thing I found while testing: Until wallet is unlocked nothing will be thrown. @DetonateYin2077 Can you check the `connectWallet` method once again? Do you have the "connect" method or similar that unlocks the wallet before requesting the account and switching to signet?
> 
> ```ts
> await this.provider?.requestAccounts() // Connect to Bitget Wallet extension
> await this.provider?.switchNetwork('signet')
> ```
> 
> Since it's not done automatically by calling this methods https://youtu.be/mzvkhGc7p88

@gbarkhatov 
The `requestAccounts` method connects to BGW. If the user hasn’t unlocked the wallet, it will display the unlock UI and then proceed to the connect authorization UI once the wallet is unlocked. The entire process is automatic and intuitive, requiring no extra actions.

Regarding the issues shown in your video, we have checked and were unable to reproduce the problem. Could you please try using the latest internal version of BGW we shared on TG this afternoon and test it again?

gbarkhatov:
@DetonateYin2077 if you disable then enable the wallet, it will be inside the locked state, right? You won't receive the issue that are recorded on the video using latest chrome extension wallet? I am using `2.9.0`

gbarkhatov (approved):

PR #150, opened 2024-05-21
fix: Activate the form one height prior the activation height

vitsalis:


jeremy-babylonchain (approved):

jrwbabylonchain:
Agree with the change in this PR. But I would like to add a follow up ticket in backlog to clean up all those logic quite a bit.
- We shall fetch global param and height at top react layer. then pass down both of it as global context/react props.
- We shall pass the full list of global params with versions. and let the individual react component to decide which versioned param it shall use by calling the utils method we already have.

React performance should not be a concern at all as:
1. Global param rarely change. so no re-rendering will happen
2. Height only change once per 10 minutes etc. not a concern too

PR #151, opened 2024-05-21
feature: Wallet select

### LINKED ISSUE #143: Wallet connect
We should be able to support multiple BTC wallets
- UI changes
- Logic changes
- Wallet example

gbarkhatov:
Adding a list of wallets `src/utils/wallet/list.ts`

To have an integration, the wallet partner needs to

1. Add:
```ts
import { OKXWallet } from "./providers/okx_wallet";
import okxIcon from "./icons/okx.svg";
import { AnotherWallet } from "./providers/another_wallet";
import anotherWalletIcon from "./icons/anotherWallet.svg";

export const walletList = [
  {
    name: "OKX",
    icon: okxIcon,
    wallet: OKXWallet,
  },
  {
    name: "Another Wallet",
    icon: anotherWalletIcon,
    wallet: AnotherWallet,
  },
];
```
2. Add a wallet type to window: `src/utils/types.ts`
2. Implement `AnotherWallet` class

Closes #143 

---

@vitsalis what approach should we take with the previously `default` wallet (injecting on mobile)? We had no selection, only a button "connect" that was leading us to `getWallet` which was choosing `window.btcwallet` or `OKX`.

Should we create a wallet class similar to `OKX` with bindings to `window.btcwallet`?


vitsalis:
> @vitsalis what approach should we take with the previously default wallet (injecting on mobile)? We had no selection, only a button "connect" that was leading us to getWallet which was choosing window.btcwallet or OKX.
> Should we create a wallet class similar to OKX with bindings to window.btcwallet?

How about if the `window.btcwallet` is already set, we do not display the wallet list and just connect the wallet under `window.btcwallet`? If we are on browser, it should not be set until the user selects a wallet. Basically the existence of `window.btcwallet` describes whether there's an available wallet or not.

vitsalis:
Let's keep the same functionality we had as before (i.e. what I described) and explore how we can improve the overall UX in a separate issue? I think this is not fully related to this PR (as it is about wallet connect)

gbarkhatov:
@vitsalis updated

vitsalis (approved):

PR #164, opened 2024-05-22
feature: Wallet select additions

gbarkhatov:
- Solves an issue with browsers using `window.btcwallet`
- Wallets should be added to `src/utils/wallet/list.ts`
- Implement an interface similar to `src/utils/wallet/providers/okx_wallet.ts` with `okxProvider`. `okxProvider` defines what window object we are calling. Like `window.okxwallet`
- Wallets extensions that are not installed but integrated will be shown as a button that leads to the docs
- If there's something in `window.btcwallet` (injection mechanism on mobile for example) we show `Browser` button that will use `window.btcwallet`

### DIFF
--- a/src/app/components/Modals/ConnectModal.tsx
+++ b/src/app/components/Modals/ConnectModal.tsx
@@ -1,4 +1,4 @@
-import { useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { Modal } from "react-responsive-modal";
 import { IoMdClose } from "react-icons/io";
 import { PiWalletBold } from "react-icons/pi";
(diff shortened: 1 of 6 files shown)


src/app/components/Modals/ConnectModal.tsx
```diff
@@ -23,16 +24,40 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
...
-      const walletProvider = walletList.find(
-        (w) => w.name === selectedWallet,
-      )?.wallet;
-      if (!walletProvider) {
-        throw new Error("Wallet provider not found");
+      let walletInstance: WalletProvider;
+
+      if (isInjectable && selectedWallet === BROWSER) {
+        // we are using the browser wallet
+        walletInstance = window[BROWSER];
```
vitsalis, line 49:
```suggestion
      if (selectedWallet === BROWSER) {
        if (!isInjectable) {
            throw new Error("Browser selected without an injectable interface").
        }
        // we are using the browser wallet
        walletInstance = window[BROWSER];
```

vitsalis (approved):

Issue #101, opened 2024-05-19
Create an example env file


supertong:
Similar to what other services we have, we always have an `env.example` file for user to copy. This is a good convention and I think we shall stick to it.

Goal: Create an `env.example` file with the example config. The example config can be copied from README file. And then update the README file to reflect the latest change.

PR #169, opened 2024-05-23
feat: add .env.example and update README

hairtail:
This PR implements feature introduced in feature request https://github.com/babylonchain/simple-staking/issues/101


vitsalis (approved):

env.example
```diff
@@ -0,0 +1,2 @@
+NEXT_PUBLIC_MEMPOOL_API=https://babylon.mempool.space
```
vitsalis, line 1: How about we rename this file to `.env.example`?

vitsalis (changes requested):

env.example
hairtail: Done

vitsalis (approved):

vitsalis:
Thanks!

PR #170, opened 2024-05-23
feature: QR wallets support

gbarkhatov:
- Adds support for QR wallets (`isQRWallet` boolean field)
- Shows a tooltip on them
- Provides an external link documentation https://i.imgur.com/s1Zf4mv.png

### DIFF
--- a/src/app/components/Modals/ConnectModal.tsx
+++ b/src/app/components/Modals/ConnectModal.tsx
@@ -4,6 +4,8 @@ import { IoMdClose } from "react-icons/io";
 import { PiWalletBold } from "react-icons/pi";
 import Image from "next/image";
 import { FaWallet } from "react-icons/fa";
+import { AiOutlineInfoCircle } from "react-icons/ai";
+import { Tooltip } from "react-tooltip";
 
 import { walletList } from "@/utils/wallet/list";
 import { WalletProvider } from "@/utils/wallet/wallet_provider";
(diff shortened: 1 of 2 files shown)


src/app/components/Modals/ConnectModal.tsx
```diff
@@ -112,31 +114,51 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
...
+                            data-tooltip-content="QR codes used for connection/signing"
+                            data-tooltip-place="top"
+                          >
+                            <AiOutlineInfoCircle />
+                          </span>
+                          <Tooltip id={name} />
+                        </div>
+                      )}
+                    </div>
+                    {isQRWallet && (
```
vitsalis, line 148: Why add this additional documentation field? Can't we do the same as with other wallets, i.e. the `href` above?

gbarkhatov: Usual wallets
- Check if it's inside the `window.provider`
- Yes - allow button to be selectable
- No - point to the documentation

Integrated QR code wallets cannot be checked with `window.provider`
- So if `clicking` on this button will always lead to a link, how can the user select it.
- If not - allow to select without the docs - how can the user check the docs?

vitsalis: Aha I see, so no need to point to documentation, the documentation pop-up is basically to allow users to install the wallet.

gbarkhatov: >Aha I see, so no need to point to documentation, the documentation pop-up is basically to allow users to install the wallet.
Removed

vitsalis (approved):

PR #171, opened 2024-05-24
fix text wrap issue on summary

jeremy-babylonchain:


src/app/components/Staking/Form/StakingAmount.tsx
```diff
@@ -87,12 +87,12 @@ export const StakingAmount: React.FC<StakingAmountProps> = ({
     } else {
       setError("");
       onStakingAmountSatChange(satoshis);
-      setValue(maxDecimals(satoshiToBtc(satoshis), 8).toString());
+      setValue(maxDecimals(satoshiToBtc(satoshis), 6).toString());
```
vitsalis, line 90: Here this should not happen. Bitcoin has 8 decimals, as the lowest unit satosh = 10^(-8) Bitcoin

vitsalis (changes requested):
Why this change?

src/app/components/Staking/Form/StakingAmount.tsx
jeremy-babylonchain: yes we should only set display value, my bad

vitsalis:
In general, why are we performing this update? What issue have we encountered?

PR #173, opened 2024-05-24
fix first activation height

jeremy-babylonchain:
return firstActivationHeight from getCurrentGlobalParamsVersion so it shows the height before the activation block

### DIFF
--- a/src/app/components/Delegations/Delegations.tsx
+++ b/src/app/components/Delegations/Delegations.tsx
@@ -111,6 +111,10 @@ export const Delegations: React.FC<DelegationsProps> = ({
         paramVersions,
       );
 
+    if (!globalParamsWhenStaking) {
+      throw new Error("Current version not found");
+    }
+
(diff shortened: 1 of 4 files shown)


jrwbabylonchain:
LGTM. Could u attach a screenshot in the PR? Thanks

Just for the purpose of notes here:
We have discussed a better approach to handle this, which is to remove the complex business logic around the `useQuery` in the page.tsx when fetching height and globa params.
Instead, those shall be the global values (context) that can be passed down to other react component. We then use the exisiting util method to get the current param versino or next param version depends on the use case.
However, this involves quite some changes and we don't think it's the right time to do it.
It's already being capture as an issue in the backlog

jeremy-babylonchain:
Screenshot 2024-05-25 at 18 06 36

gbarkhatov (approved):

jrwbabylonchain (approved):

PR #181, opened 2024-05-25
fix - inconsistent delegation status 

jeremy-babylonchain:


src/app/components/Delegations/Delegations.tsx
```diff
@@ -152,7 +152,7 @@ export const Delegations: React.FC<DelegationsProps> = ({
     const stakerSignature = unbondingTx.ins[0].witness[0].toString("hex");
 
     // POST unbonding to the API
-    postUnbonding(
+    await postUnbonding(
```
gbarkhatov, line 155: How did we loose that await here? 🫤

src/app/components/Delegations/Delegations.tsx
```diff
@@ -338,8 +338,18 @@ export const Delegations: React.FC<DelegationsProps> = ({
 
   // Combine the delegations from the API and local storage
   const combinedDelegationsData = delegationsAPI
-    ? [...delegationsLocalStorage, ...delegationsAPI]
-    : delegationsLocalStorage;
+  ? [
+      ...delegationsAPI,
+      ...delegationsLocalStorage.filter(
```
gbarkhatov, line 343: Can we make a comment here about whats happening

gbarkhatov (approved):

src/app/components/Delegations/Delegations.tsx
```diff
@@ -328,18 +326,31 @@ export const Delegations: React.FC<DelegationsProps> = ({
...
   }, [delegationsAPI, setIntermediateDelegationsLocalStorage]);
 
-  // Combine the delegations from the API and local storage
+  // combine delegations from the API and local storage, prioritizing API data
   const combinedDelegationsData = delegationsAPI
-    ? [...delegationsLocalStorage, ...delegationsAPI]
-    : delegationsLocalStorage;
+    ? [
+        ...delegationsAPI,
+        // filter local storage delegations to EXCLUDE those already present in the API
```
vitsalis, line 341: I wonder why we need this though. Don't we have an effect that deletes delegations that are on the API from the local storage? If we don't delete them, this might lead to the app silently consuming more and more local storage

jeremy-babylonchain: agreed and resolved

vitsalis (approved):

gbarkhatov:
@jeremy-babylonchain intermediate delegations in local storage are cleaned up when we withdraw and switch wallet addresses. Maybe we should have a condition?
- for intermediate unbonded transactions (local storage) we check if API delegation has:
  - UNBONDING_REQUESTED
  - UNBONDING
  - UNBONDED
- for intermediate withdrawn transactions (local storage) we check if API delegation has
  - WITHDRAWN

I.e. we remove unbonding intermediate local storage delegation if it's requested, unbonding or unbonded. Withdrawn delegation should be removed only if we receive withdrawn status

So `INTERMEDIATE_WITHDRAWAL` local storage items are not removed when we have `unbonding` API state

gbarkhatov (approved):

PR #185, opened 2024-05-26
Adopt dynamic calculated fees

jrwbabylonchain:
Pending merge of this PR https://github.com/babylonchain/btc-staking-ts/pull/24 then we can bump the ts lib version here

But the implementation needed in this repo is pretty much ready for review

### DIFF
--- a/src/app/components/Delegations/Delegations.tsx
+++ b/src/app/components/Delegations/Delegations.tsx
@@ -2,6 +2,7 @@ import { useEffect, useState } from "react";
 import { useLocalStorage } from "usehooks-ts";
 import { Transaction, networks } from "bitcoinjs-lib";
 import {
+  PsbtTransactionResult,
   unbondingTransaction,
   withdrawEarlyUnbondedTransaction,
   withdrawTimelockUnbondedTransaction,
@@ -232,10 +233,10 @@ export const Delegations: React.FC<DelegationsProps> = ({
     );
 
     // Create the withdrawal transaction
-    let transactionResult;
+    let withdrawPsbtTxResult: PsbtTransactionResult;
     if (delegation?.unbondingTx) {
       // Withdraw funds from an unbonding transaction that was submitted for early unbonding and the unbonding period has passed
-      transactionResult = withdrawEarlyUnbondedTransaction(
+      withdrawPsbtTxResult = withdrawEarlyUnbondedTransaction(
         {
           unbondingTimelockScript,
           slashingScript,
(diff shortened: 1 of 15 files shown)


jeremy-babylonchain (approved):

jeremy-babylonchain:
looks good to me. will help test again when ts lib is ready

src/app/components/Delegations/Delegations.tsx
```diff
@@ -218,57 +212,59 @@ export const Delegations: React.FC<DelegationsProps> = ({
...
         delegation.stakingTx.outputIndex,
       );
     }
 
     // Sign the withdrawal transaction
     let withdrawalTransaction: Transaction;
     try {
-      withdrawalTransaction = await signPsbtTx(unsignedWithdrawalTx.toHex());
+      const { psbt } = withdrawPsbtTxResult;
+      withdrawalTransaction = await signPsbtTx(psbt.toHex());
```
vitsalis, line 262:
```suggestion
      withdrawalTransaction = await signPsbtTx(withdrawPsbtTxResult.psbt.toHex());
```

src/utils/signForm.ts
```diff
@@ -53,38 +49,40 @@ export const signForm = async (
...
-      unbondingScript,
-      slashingScript,
+    const netWorkFee = await btcWallet.getNetworkFees();
+    feeRate = netWorkFee.fastestFee;
+  } catch (error) {
+    throw new Error("Cannot get network fees");
+  }
+  let unsignedStakingPsbt;
+  try {
+    const { psbt } = stakingTransaction(
```
vitsalis, line 61:
```suggestion
    unsignedStakingPsbt = stakingTransaction(
```

vitsalis (approved):
lgtm

src/utils/signForm.ts
gbarkhatov: @vitsalis it returns
```ts
PsbtTransactionResult {
    psbt: Psbt;
    fee: number;
}
```

gbarkhatov (commented):
When we click stake we produce a lot of mempool calls https://youtu.be/Vj-dAtCcl0M
1. Is this what we want?
In any case we need to
1. Disable the "stake" button, until we get the response from the wallet UI
2. Show a loading indicator instead of the text

vitsalis:
> When we click stake we produce a lot of mempool calls https://youtu.be/Vj-dAtCcl0M
> 
>     1. Is this what we want?
>        In any case we need to

Indeed, that's not nice. It happens because we get * all * UTXOs and for each one we query to get additional information about them. A neat solution to this would be to first get the basic information about all UTXOs and then query the additional information only for those that we selected.

jrwbabylonchain:
> When we click stake we produce a lot of mempool calls https://youtu.be/Vj-dAtCcl0M
> 
> 1. Is this what we want?
>    In any case we need to
> 2. Disable the "stake" button, until we get the response from the wallet UI
> 3. Show a loading indicator instead of the text

resolved in the latest commit by using the `validate-address` endpoint

PR #209, opened 2024-05-28
fix: fix typo and displaying sBTC instead of BTC

jrwbabylonchain:

### DIFF
--- a/src/app/components/Staking/Staking.tsx
+++ b/src/app/components/Staking/Staking.tsx
@@ -235,7 +235,7 @@ export const Staking: React.FC<StakingProps> = ({
         <Message
           title="Staking has not yet started"
           messages={[
-            `Staking will be activated at Signet BTC block height ${firstActivationHeight || "-"}. The current Signet BTC block height is ${height || "-"}.`,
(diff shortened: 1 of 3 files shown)


src/app/components/Staking/Staking.tsx
```diff
@@ -235,8 +235,7 @@ export const Staking: React.FC<StakingProps> = ({
         <Message
           title="Staking has not yet started"
           messages={[
-            `The staking application will open once the next Bitcoin block height is the activation Bitcoin block height (${firstActivationHeight || "-"}).`,
-            `The next Bitcoin block height is ${height || "-"}.`,
+            `Staking will be activated at Signet BTC block height ${firstActivationHeight || "-"}. The current Signet BTC block height is ${height || "-"}.`,
```
jeremy-babylonchain, line 238:
```suggestion
            `Staking will be activated at ${coinName} block height ${firstActivationHeight || "-"}. The current ${coinName} block height is ${height || "-"}.`,
```

jrwbabylonchain: i did it on purpose as we are so close to have it turned on. we can change it to dynamic value later on.

vitsalis (approved):

PR #210, opened 2024-05-28
fix: Connect modal injectable key

gbarkhatov:
`Warning: Each child in a list should have a unique "key" prop.`
Add a key on injectable wallet


gbarkhatov:
Closed as covered in #205

PR #213, opened 2024-05-29
display staking cap before wallet login

jrwbabylonchain:
image

Ideally the global param and height fetch in the page.tsx can be improved using same/similar method.
But since we are currently at a testing stage. we can leave it for now and deal with it later on

### DIFF
--- a/src/app/context/api/GlobalParamsProvider.tsx
+++ b/src/app/context/api/GlobalParamsProvider.tsx
@@ -1,0 +1,34 @@
+import React, { ReactNode, createContext, useContext } from "react";
+import { useQuery } from "@tanstack/react-query";
+import { getTipHeight } from "@/utils/mempool_api";
+import { getGlobalParams } from "@/app/api/getGlobalParams";
+import { GlobalParamsVersion } from "@/app/types/globalParams";
+
+interface GlobalParamsProviderProps {
+  children: ReactNode;
+}
+
+const GlobalParamsContext = createContext<GlobalParamsVersion[] | undefined>(
+  undefined,
+);
+
+export const GlobalParamsProvider: React.FC<GlobalParamsProviderProps> = ({
+  children,
+}) => {
+  const { data } = useQuery({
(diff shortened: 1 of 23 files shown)


src/app/context/api/GlobalParamsProvider.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { ReactNode, createContext, useContext } from "react";
+import { useQuery } from "@tanstack/react-query";
+import { getTipHeight } from "@/utils/mempool_api";
```
gbarkhatov, line 3: should we separate package imports from the local imports?

src/app/context/api/GlobalParamsProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const GlobalParamsContext = createContext<GlobalParamsVersion[] | undefined>(
+  undefined,
+);
+
+export const GlobalParamsProvider: React.FC<GlobalParamsProviderProps> = ({
+  children,
+}) => {
+  const { data } = useQuery({
+    queryKey: ["API_GLOBAL_PARAMS"],
```
gbarkhatov, line 19: I think we need to either uppercase all of them, or lowercase all of them:
https://i.imgur.com/fo9t50R.png

gbarkhatov (approved):

src/app/context/api/GlobalParamsProvider.tsx
jrwbabylonchain: yeah. generally speaking, i prefer uppercase and no space (use `_`) for any key identifier.
If you agree with this approach, we can have a low priority task to change all existing ones to this.
But again, i don't have strong opinion on this. If not bothered, happy to change back to the current format which i think is `api global params`

src/app/context/api/GlobalParamsProvider.tsx
jrwbabylonchain: probably a good way to handle this to let the npx format to fix it.

PR #225, opened 2024-06-03
chore: change Staker's type definition to camel case

LouisWT:


gbarkhatov:
@LouisWT 
Thank you so much for your contribution! Our code is currently frozen for the security audit. We will review and adapt your PR once it is safe to do so. Thanks again!

PR #285, opened 2024-06-24
fix: Bitget remove support

gbarkhatov:
At the moment Bitget does not support unbonding or withdrawal
Estimated update - June 26th


gbarkhatov:
Closing as it works now

PR #286, opened 2024-06-25
feature: Test updateDelegations

gbarkhatov:
- `updateDelegations` is a pure function
- tests for :
  - different
  - same
  - empty
  - identical

Closes #282

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -8,8 +8,8 @@ import { useLocalStorage } from "usehooks-ts";
 
 import { network } from "@/config/network.config";
 import { getCurrentGlobalParamsVersion } from "@/utils/globalParams";
-import { filterDelegationsLocalStorage } from "@/utils/local_storage/filterDelegationsLocalStorage";
 import { getDelegationsLocalStorageKey } from "@/utils/local_storage/getDelegationsLocalStorageKey";
+import { updateDelegations } from "@/utils/local_storage/updateDelegations";
 import { WalletError, WalletErrorType } from "@/utils/wallet/errors";
 import {
   getPublicKeyNoCoord,
@@ -287,36 +287,11 @@ const Home: React.FC<HomeProps> = () => {
       return;
     }
 
-    const updateDelegations = async () => {
-      // Filter the delegations that are still valid
-      const validDelegations = await filterDelegationsLocalStorage(
-        delegationsLocalStorage,
-        delegations.delegations,
-      );
-
-      // Extract the stakingTxHashHex from the validDelegations
-      const validDelegationsHashes = validDelegations
-        .map((delegation) => delegation.stakingTxHashHex)
-        .sort();
-      const delegationsLocalStorageHashes = delegationsLocalStorage
-        .map((delegation) => delegation.stakingTxHashHex)
-        .sort();
-
-      // Check if the validDelegations are different from the current delegationsLocalStorage
-      const areDelegationsDifferent =
-        validDelegationsHashes.length !==
-          delegationsLocalStorageHashes.length ||
-        validDelegationsHashes.some(
-          (hash, index) => hash !== delegationsLocalStorageHashes[index],
-        );
-
-      // Update the local storage delegations if they are different to avoid unnecessary updates
-      if (areDelegationsDifferent) {
-        setDelegationsLocalStorage(validDelegations);
-      }
-    };
-
-    updateDelegations();
+    updateDelegations(
+      delegations.delegations,
+      delegationsLocalStorage,
+      setDelegationsLocalStorage,
+    );
   }, [delegations, setDelegationsLocalStorage, delegationsLocalStorage]);
 
   // Finality providers key-value map { pk: moniker }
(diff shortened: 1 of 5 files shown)


src/utils/local_storage/updateDelegations.ts
```diff
@@ -0,0 +1,37 @@
+import { Delegation } from "@/app/types/delegations";
+
+import { filterDelegationsLocalStorage } from "./filterDelegationsLocalStorage";
+
+export const updateDelegations = async (
```
vitsalis, line 5:
```suggestion
export const calculateDelegationsDiff = async (
```

This does not do any updating.

src/utils/local_storage/updateDelegations.ts
```diff
@@ -0,0 +1,37 @@
...
+  delegations: Delegation[],
+  delegationsLocalStorage: Delegation[],
+): Promise<{ areDelegationsDifferent: boolean; delegations: Delegation[] }> => {
+  // Filter the delegations that are still valid
+  const validDelegations = await filterDelegationsLocalStorage(
```
vitsalis, line 10:
```suggestion
  const validLocalStorageDelegations = await filterDelegationsLocalStorage(
```

src/utils/local_storage/updateDelegations.ts
```diff
@@ -0,0 +1,37 @@
...
+      (hash, index) => hash !== delegationsLocalStorageHashes[index],
+    );
+
+  return {
+    areDelegationsDifferent,
+    // Return the new delegations if they are different or original if no update is needed
+    delegations: areDelegationsDifferent
+      ? validDelegations
+      : delegationsLocalStorage,
+  };
```
vitsalis, line 36: Don't understand why we are doing the above diff between `validDelegations` and `delegationsLocalStorage`. If they are different, we return `validDelegations`. If they are the same, we return `delegationsLocalStorage`. Why not return `validDelegations` at all times? Won't we get the same result?

If we are doing this just to know whether we should call `setDelegationsLocalStorage`, then (1) is the overhead of `setDelegationsLocalStorage` worth the additional complexity and (2) this seems a bit of spaghetti code for a utility to do something random completely unrelated to its purpose for a function up the stack. If we want to do this, we should have two utilities: (1) `filterLocalStorageDelegations(delegationsLocalStorage, delegationsAPI)` and (2) `delegationsDifferent(delegations1, delegations2): bool`.

tests/utils/local_storage/updateDelegations.test.ts
```diff
@@ -0,0 +1,91 @@
...
+// Mock the filterDelegationsLocalStorage function
+jest.mock("@/utils/local_storage/filterDelegationsLocalStorage");
+
+describe("updateDelegations", () => {
+  const mockDelegationsHash1: Delegation[] = [
+    { stakingTxHashHex: "hash1" } as Delegation,
+  ];
+  const mockDelegationsHash2: Delegation[] = [
+    { stakingTxHashHex: "hash2" } as Delegation,
+  ];
```
vitsalis, line 14: We should have more elements in each list. Comparing one by one does not cover a lot of test cases. I recommend we have multiple lists, e.g. some with 1 element, others with many etc.

tests/utils/local_storage/updateDelegations.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  it("should handle both empty delegations and local storage correctly", async () => {
+    (filterDelegationsLocalStorage as jest.Mock).mockResolvedValue([]);
+
+    const result = await updateDelegations([], []);
+
+    expect(result.areDelegationsDifferent).toBe(false);
+    expect(result.delegations).toEqual([]);
+  });
+
+  it("should handle identical delegations correctly", async () => {
```
vitsalis, line 78: This test has already been performed.

src/utils/local_storage/updateDelegations.ts
gbarkhatov: Done

src/utils/local_storage/updateDelegations.ts
gbarkhatov: Done

tests/utils/local_storage/updateDelegations.test.ts
gbarkhatov: Done

src/utils/local_storage/updateDelegations.ts
gbarkhatov: Done

vitsalis (approved):

src/app/page.tsx
```diff
@@ -287,36 +287,18 @@ const Home: React.FC<HomeProps> = () => {
...
-        .map((delegation) => delegation.stakingTxHashHex)
-        .sort();
-
-      // Check if the validDelegations are different from the current delegationsLocalStorage
-      const areDelegationsDifferent =
-        validDelegationsHashes.length !==
-          delegationsLocalStorageHashes.length ||
-        validDelegationsHashes.some(
-          (hash, index) => hash !== delegationsLocalStorageHashes[index],
+    const update = async () => {
```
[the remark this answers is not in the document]
gbarkhatov: Okay

PR #321, opened 2024-07-03
feat: add imToken wallet provider

huiliu1110:
Integrate imToken wallet


gbarkhatov:
If `imToken` uses mobile injectable interface then this PR is not needed

vitsalis:
Closing as imToken is a mobile wallet, while this functionality is only needed for browser extension wallets.

PR #322, opened 2024-07-03
chores: add unit test for sign psbt transaction 

### LINKED ISSUE #280: Unit test for `signPsbtTransaction`
The method `signPsbtTransaction` is a util method but not inside the utils directory. We will consider move under utils at later stage, but for now we will need unit tests to cover this method

jeremy-babylonchain:
Closes #280


gbarkhatov (approved):
Looks good after @jrwbabylonchain comments

PR #347, opened 2024-07-12
hotfix - 337 navbar balance and address fix

### LINKED ISSUE #337: Navbar balance and address in mobile overlap
image

jeremy-babylonchain:
Closes #337

### DIFF
--- a/src/app/components/NetworkBadge/NetworkBadge.tsx
+++ b/src/app/components/NetworkBadge/NetworkBadge.tsx
@@ -10,7 +10,7 @@ interface NetworkBadgeProps {}
 
 export const NetworkBadge: React.FC<NetworkBadgeProps> = () => {
   return (
-    <div className="absolute left-2 top-[6rem]">
+    <div className="absolute left-2 top-[10rem] md:top-[6rem]">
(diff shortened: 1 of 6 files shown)


src/app/components/NetworkBadge/NetworkBadge.tsx
gbarkhatov, line 1: - On this breakpoint the Network Badge is "inside" the header part: https://i.imgur.com/vuevXhy.png
- I think it should have different positions when connected/not connected: https://i.imgur.com/srm4SD6.png because right now it looks okay when we're connected, but too much when we're not. Even though I don't like the whole idea of this badge floating around there at all, but from our previous discussions we should keep it there. You can use `style` based on prop and set the top parameter, instead of Tailwind approach

gbarkhatov (changes requested):

gbarkhatov:
@jeremy-babylonchain please ping me out when all the changes are applied and we are sure about what's the design that we want (for that particular case and component)

jeremy-babylonchain:
https://github.com/user-attachments/assets/a179a80c-5701-4e0a-8024-b7929e46ab32

gbarkhatov (approved):

PR #381, opened 2024-07-25
feat: Cronjob to sync from main to dev

hiepmai-babylonchain:


hiepmai-babylonchain:
The cron only gets triggered on the default branch so that's why the PR is on `dev`

gbarkhatov:
I am not sure I can add anything to this infra related PR

PR #382, opened 2024-07-25
feature: Geo blocking with 451

gbarkhatov:
- general health check error: https://i.imgur.com/ozubBzB.png
- geo 451 error: https://i.imgur.com/qQOScR4.png

### DIFF
--- a/src/app/api/apiWrapper.ts
+++ b/src/app/api/apiWrapper.ts
@@ -35,6 +35,11 @@ export const apiWrapper = async (
     );
   } catch (error) {
     if (axios.isAxiosError(error)) {
+      // 451 is a custom status code for geo-blocking
+      if (error.response?.status === 451 || error.request.status === 451) {
+        // Rethrow the original error so the caller can handle it
+        throw error;
+      }
       const message = error?.response?.data?.message;
       throw new Error(message || generalErrorMessage);
     } else {
--- a/src/app/api/getHealthCheck.ts
+++ b/src/app/api/getHealthCheck.ts
@@ -1,0 +1,44 @@
+import { isAxiosError } from "axios";
+
+import { HealthCheckResult } from "../types/healthCheck";
+
+import { apiWrapper } from "./apiWrapper";
+
+interface HealthCheckResponse {
+  data: string;
+}
+
+export const getHealthCheck = async (): Promise<HealthCheckResult> => {
+  try {
+    const response = await apiWrapper(
+      "GET",
+      "/healthcheck",
+      "Error checking health",
+    );
+    const healthCheckAPIResponse: HealthCheckResponse = response.data;
+    // If the response has a data field, it's a normal response
+    if (healthCheckAPIResponse.data) {
+      return { status: "normal", message: healthCheckAPIResponse.data };
+    } else {
+      // Something went wrong
+      throw new Error("Health check failed");
+    }
+  } catch (error: Error | any) {
+    // Geo-blocking is a custom status code
+    if (
+      isAxiosError(error) &&
+      (error.response?.status === 451 || error.request.status === 451)
+    ) {
+      return {
+        status: "geoBlocked",
+        message:
+          "The Bitcoin Staking functionality is not available in your region",
+      };
+    } else {
+      return {
+        status: "error",
+        message: "Please try again later",
+      };
+    }
+  }
+};
--- a/src/app/components/Connect/ConnectSmall.tsx
+++ b/src/app/components/Connect/ConnectSmall.tsx
@@ -1,9 +1,12 @@
 import { useRef, useState } from "react";
+import { AiOutlineInfoCircle } from "react-icons/ai";
 import { FaBitcoin } from "react-icons/fa";
 import { IoMdClose } from "react-icons/io";
 import { PiWalletBold } from "react-icons/pi";
+import { Tooltip } from "react-tooltip";
 import { useOnClickOutside } from "usehooks-ts";
 
+import { HealthCheckResult } from "@/app/types/healthCheck";
 import { getNetworkConfig } from "@/config/network.config";
 import { satoshiToBtc } from "@/utils/btcConversions";
 import { maxDecimals } from "@/utils/maxDecimals";
@@ -17,6 +20,7 @@ interface ConnectSmallProps {
   address: string;
   btcWalletBalanceSat?: number;
   onDisconnect: () => void;
+  healthCheck?: HealthCheckResult;
 }
 
 export const ConnectSmall: React.FC<ConnectSmallProps> = ({
@@ -24,6 +28,7 @@ export const ConnectSmall: React.FC<ConnectSmallProps> = ({
   address,
   btcWalletBalanceSat,
   onDisconnect,
+  healthCheck,
 }) => {
   const [showMenu, setShowMenu] = useState(false);
   const handleClickOutside = () => {
@@ -34,6 +39,26 @@ export const ConnectSmall: React.FC<ConnectSmallProps> = ({
   useOnClickOutside(ref, handleClickOutside);
 
   const { coinName, networkName } = getNetworkConfig();
+
+  const renderHealthCheckTooltip = () => {
+    if (!healthCheck) return null;
+
+    return (
+      <>
+        <span
+          className="cursor-pointer text-xs"
+          data-tooltip-id={`tooltip-connect-${healthCheck.status}`}
+          data-tooltip-content={healthCheck.message}
+          data-tooltip-place="bottom"
+        >
+          <AiOutlineInfoCircle />
+        </span>
+        <Tooltip id={`tooltip-connect-${healthCheck.status}`} />
+      </>
+    );
+  };
+
+  const healthCheckNormal = healthCheck?.status === "normal";
 
   return address ? (
     <div className="relative mr-[-10px] text-sm hidden md:flex" ref={ref}>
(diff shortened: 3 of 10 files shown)


src/app/api/apiWrapper.ts
```diff
@@ -35,6 +35,11 @@ export const apiWrapper = async (
     );
   } catch (error) {
     if (axios.isAxiosError(error)) {
+      // 451 is a custom status code for geo-blocking
+      if (error.response?.status === 451 || error.request.status === 451) {
```
vitsalis, line 39: This seems a bit hacky. Why are we doing this just for 451? If we are doing this so that we can use `apiWrapper` for the `healthcheck` endpoint, let's just not use this wrapper for the healthcheck endpoint and directly do the request from the method.

src/app/api/getHealthCheck.ts
```diff
@@ -0,0 +1,44 @@
...
+      throw new Error("Health check failed");
+    }
+  } catch (error: Error | any) {
+    // Geo-blocking is a custom status code
+    if (
+      isAxiosError(error) &&
+      (error.response?.status === 451 || error.request.status === 451)
+    ) {
+      return {
+        status: "geoBlocked",
```
vitsalis, line 33:
```suggestion
        status: "geoblocked",
```

Weird for a status to have camel case

src/app/components/Connect/ConnectSmall.tsx
```diff
@@ -35,6 +40,26 @@ export const ConnectSmall: React.FC<ConnectSmallProps> = ({
 
   const { coinName, networkName } = getNetworkConfig();
 
+  const renderHealthCheckTooltip = () => {
```
vitsalis, line 43: This tooltip is not about `healthcheck`, it's about showing whether someone is geo-blocked or not.

src/app/components/Connect/ConnectSmall.tsx
```diff
@@ -35,6 +40,26 @@ export const ConnectSmall: React.FC<ConnectSmallProps> = ({
...
+    if (!healthCheck) return null;
+
+    return (
+      <>
+        <span
+          className="cursor-pointer text-xs"
+          data-tooltip-id={`tooltip-connect-${healthCheck.status}`}
+          data-tooltip-content={healthCheck.message}
```
vitsalis, line 51: Why are we getting the content for a tooltip directly out of the error returned from an API response? The API response should not specify the content, the API response should specify what went wrong, and based on what went wrong we construct the content in the content component.

src/app/types/healthCheck.ts
```diff
@@ -0,0 +1,4 @@
+export type HealthCheckResult =
+  | { status: "normal"; message: string }
```
vitsalis, line 2: `normal`, `geoblocked` and `error` should be constants. 

Otherwise, we could also have an enum, but I understand that's difficult in js.

src/app/components/Staking/Staking.tsx
```diff
@@ -484,15 +489,39 @@ export const Staking: React.FC<StakingProps> = ({
...
-    if (!isWalletConnected) {
+    // Health check failed
+    if (!healthCheck || healthCheck.status === "error") {
+      return (
+        <Message
+          title="Staking is not available"
+          messages={[
+            "The Bitcoin Staking functionality is not available at the moment.",
+          ]}
+          icon={healthCheckFailed}
```
vitsalis, line 500: This icon has nothing to do with health check. Health check is the endpoint that we use to check whether the API is available. The icon name should be something along the lines of `apiNotAvailable`

src/app/components/Staking/Staking.tsx
```diff
@@ -83,6 +87,7 @@ export const Staking: React.FC<StakingProps> = ({
   setDelegationsLocalStorage,
   btcWalletBalanceSat,
   availableUTXOs,
+  healthCheck,
```
vitsalis, line 90:
```suggestion
  apiAvailable,
```

vitsalis (changes requested):
We need to have a separation between the internal representation of the API being available and the fact that we are using the health check endpoint to accomplish that.

src/app/api/apiWrapper.ts
gbarkhatov: If we do it here we can rethrow an `AxiosError`. If we don't do it and provide our own message there's no way to separate what the error was as
- `451`
- And `404` for example
all of the above will have the same error with the same message

gbarkhatov: >let's just not use this wrapper for the healthcheck endpoint and directly do the request from the method.
Okay

src/app/components/Connect/ConnectSmall.tsx
gbarkhatov: It is health check indeeed, if we get anything besides the normal response we will render an error text OR geo blocking text

src/app/components/Connect/ConnectSmall.tsx
gbarkhatov: Ok

src/app/components/Staking/Staking.tsx
gbarkhatov: @vitsalis you want it only in Staking? Or do you want me to change the `healthCheck` completely to `apiAvailable`?

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/app/components/Staking/Staking.tsx
gbarkhatov: Done

src/app/types/healthCheck.ts
gbarkhatov: Done

src/app/components/Connect/ConnectSmall.tsx
gbarkhatov: Done

src/app/components/Connect/ConnectSmall.tsx
gbarkhatov: Done

src/app/api/getHealthCheck.ts
gbarkhatov: Done

src/app/api/apiWrapper.ts
gbarkhatov: Done

src/app/components/Connect/ConnectSmall.tsx
```diff
@@ -17,13 +25,15 @@ interface ConnectSmallProps {
   address: string;
   btcWalletBalanceSat?: number;
   onDisconnect: () => void;
+  apiAvailable?: HealthCheckResult;
```
vitsalis, line 28: Why is this `?`?

src/app/components/Connect/ConnectSmall.tsx
```diff
@@ -35,6 +45,31 @@ export const ConnectSmall: React.FC<ConnectSmallProps> = ({
 
   const { coinName, networkName } = getNetworkConfig();
 
+  const renderApiAvailableTooltip = () => {
```
vitsalis, line 48:
```suggestion
  const renderApiNotAvailableTooltip = () => {
```

vitsalis: Let's also have a comment describing what this function's purpose is.

src/app/components/Connect/ConnectSmall.tsx
```diff
@@ -92,13 +127,16 @@ export const ConnectSmall: React.FC<ConnectSmallProps> = ({
...
-      disabled={!!address}
-    >
-      <PiWalletBold size={20} className="flex md:hidden" />
-      <span className="hidden md:flex">Connect to {networkName} network</span>
-    </button>
+    <div className="flex items-center gap-1">
+      <button
+        className="btn-primary btn h-[2.5rem] min-h-[2.5rem] rounded-full px-2 text-white md:rounded-lg"
+        onClick={onConnect}
+        disabled={!!address || !isApiNormal}
```
vitsalis, line 134: Let's add a comment denoting that we disable the button if the API is not healthy.

vitsalis (approved):
Ideally, I would like an extra layer between healthcheck and the rest of the application, but this might be overengineering, so lgtm.

src/app/components/Connect/ConnectSmall.tsx
gbarkhatov: It is undefined before we fetch and get into one of the `normal - geoblocking - error` state

src/app/components/Connect/ConnectSmall.tsx
gbarkhatov: Done

src/app/components/Connect/ConnectSmall.tsx
gbarkhatov: Done

gbarkhatov:
@jrwbabylonchain @jeremy-babylonchain waiting for the 2nd pair of eyes

gbarkhatov:
@jrwbabylonlab 
1. I will check whether or not I will apply it as a more domain specific naming and placement, separation of concerns. But the same approach with "good opportunity to start making better stuff" might be applied to new additions to 700 lines of code `Staking.tsx`
2. `apiAvailable` is used in 2 places - `Header` and `Staking`. I would not put any state to global (whether it's `context` or `redux` or whatever) unless I have to. Looking at UI side, I believe at some point of time we will have only one place to connect the wallet (not we have UI connect duplication), so as later on when we refactor and improve I think we _might_ cut the amount of `context` providers. I'll move this PR to `labs`

